PR #6, opened 2021-12-09
Use shared GitHub Actions

slawekjaranowski:


slawekjaranowski:
At the moment build will not pass ... but we will have short feedback for next PR.

slachiewicz (approved):

PR #10, opened 2021-12-12
[MWRAPPER-37] restore MVNW_REPOURL support for wrapper plugin

hboutemy:
https://issues.apache.org/jira/browse/MWRAPPER-37


jira-importer:
Resolve #202

PR #12, opened 2021-12-20
[MWRAPPER-43] Make jar download quiet by default

jorsol:
This adds `--quiet` to wget and `--silent` to curl by default, and only when MVNW_VERBOSE is set to true it prints the download information.

Also the MavenWrapperDownloader.java is modified to read and print based on MVNW_VERBOSE.


hboutemy:
perfect, thank you

jira-importer:
Resolve #240

jira-importer:
Resolve #240

PR #14, opened 2022-01-23
[MWRAPPER-47] - Use name wrapperUrl consistently in Maven Wrapper files/scripts

jimisola:
Replaced jarUrl with wrapperUrl in mvnw

My first maven contribution in 15 years or so. Please advise if I missed out on something.

### DIFF
--- a/maven-wrapper-distribution/src/resources/mvnw
+++ b/maven-wrapper-distribution/src/resources/mvnw
@@ -216,16 +216,16 @@ else
       echo "Couldn't find .mvn/wrapper/maven-wrapper.jar, downloading it ..."
     fi
     if [ -n "$MVNW_REPOURL" ]; then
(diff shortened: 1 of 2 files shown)


hboutemy:
nice
while at it, we could rename also DOWNLOAD_URL to WRAPPER_URL in mvnw.cmd, WDYT?
and eventually better align choices in mvnw.cmd vs mvnw

jimisola:
I agree. I've renamed DOWNLOAD_URL to WRAPPER_URL in mvnw.cmd. Although, it was a simple find-replace it has not been tested since I don't use Windows. 

Not sure what choices you are referring to (command line option?), but that will be clearer with another Jira issue :)

hboutemy:
> Not sure what choices you are referring to (command line option?), but that will be clearer with another Jira issue :)

I have no concrete case in mind, just the idea that there may be other cases of such differences between mvnw.cmd and mvnw that could be avoided to have a better consistency

hboutemy:
thank you

jira-importer:
Resolve #238

PR #18, opened 2022-02-03
[MWRAPPER-55] add threadSafe = true

XenoAmess:
to make it not warn when run maven maven parallel build.

### DIFF
--- a/maven-wrapper-plugin/src/main/java/org/apache/maven/plugins/wrapper/WrapperMojo.java
+++ b/maven-wrapper-plugin/src/main/java/org/apache/maven/plugins/wrapper/WrapperMojo.java
@@ -60,7 +60,7 @@ import static org.apache.maven.shared.utils.logging.MessageUtils.buffer;
  * @author Robert Scholte
  * @since 3.0.0
  */
-@Mojo( name = "wrapper", aggregator = true, requiresDirectInvocation = true )
+@Mojo( name = "wrapper", aggregator = true, requiresDirectInvocation = true, threadSafe = true )
 public class WrapperMojo extends AbstractMojo
 {
     private static final String MVNW_REPOURL = "MVNW_REPOURL";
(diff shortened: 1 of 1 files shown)


jorsol:
In which scenario does this need to be run in a parallel build?

The normal usage is to just run `mvn -N wrapper:wrapper` to set up the wrapper, so I'm curious why this should be run in a build (or even declared as a plugin in the pom).

jorsol:
That's essentially how it works:

Generating the Wrapper files requires an installed version of the Maven environment on your machine, it's not weird. Thankfully, generating the initial Wrapper files is a one-time process, this means that the wrapper plugin should be run just once.

Then you commit those files to git and later executions can use the wrapper instead of the Maven environment.

XenoAmess:
> Then you commit those files to git and later executions can use the wrapper instead of the Maven environment.

but if yuo do not make the wapper goal into your local maven build chain, it is hard to update the committed shell codes when you upgrade somethings.

jorsol:
You need a "local" maven environment first to set up the `mvnw` script files, that's a one-time process, then later you simply use the wrapper scripts.

In a nutshell, you gain the following benefits:
- Standardizes a project on a given Maven version, leading to more reliable and robust builds.
- Provisioning a new Maven version to different users and execution environment (e.g. IDEs or Continuous Integration servers) is as simple as changing the Wrapper definition.

Maven Wrapper consists of 3 components:

1. [maven-wrapper](https://maven.apache.org/wrapper/maven-wrapper/), providing the `maven-wrapper.jar` that downloads, installs, and runs target Maven distribution,
2. [maven-wrapper-distribution](https://maven.apache.org/wrapper/maven-wrapper-distribution/), that provides `mvnw`/`mvnw.cmd` scripts distributions,
3. [maven-wrapper-plugin](https://maven.apache.org/wrapper/maven-wrapper-plugin/), the wrapper plugin to ease installing Wrapper into a project.

If you don't want to (or can't) use a "local" maven to set up the wrapper, then you can skip the plugin and manually download and extract the `wrapper-distribution`, for instance, you can download the `source` distribution that contains the shell scripts from:
https://repo1.maven.org/maven2/org/apache/maven/wrapper/maven-wrapper-distribution/3.1.0/maven-wrapper-distribution-3.1.0-source.zip

jira-importer:
Resolve #236

jira-importer:
Resolve #236

PR #21, opened 2022-02-05
Don't ignore dependabot branches for shared GH v2

slawekjaranowski:
From v2 for shared GitHub Action duplicate build on PR are skipped in script,
so we shouldn't ignore dependabot branches


olamy:
this is something which does not need a PR it just makes more notification noise for something so trivial....

PR #28, opened 2022-03-13
[MWRAPPER-23] Get maven user home (~/.m2) with the same logic with maven

gzm55:
New maven wrapper will guess the maven user home location via system props, system envs
and at last the default location (`~/.m2`).
This guessing order is the convention of the gradle wrapper.
For maven, user home has a fixed location, i.e. `~/.m2`.

### DIFF
--- a/maven-wrapper/src/main/java/org/apache/maven/wrapper/MavenWrapperMain.java
+++ b/maven-wrapper/src/main/java/org/apache/maven/wrapper/MavenWrapperMain.java
@@ -45,10 +45,6 @@ public class MavenWrapperMain
 
     private static final Path DEFAULT_MAVEN_USER_HOME = Paths.get( System.getProperty( "user.home" ) ).resolve( ".m2" );
 
-    public static final String MAVEN_USER_HOME_PROPERTY_KEY = "maven.user.home";
-
-    public static final String MAVEN_USER_HOME_ENV_KEY = "MAVEN_USER_HOME";
-
     public static final String MVNW_VERBOSE = "MVNW_VERBOSE";
 
     public static final String MVNW_USERNAME = "MVNW_USERNAME";
@@ -148,12 +144,6 @@ public class MavenWrapperMain
 
     private static Path mavenUserHome()
     {
-        String mavenUserHome = System.getProperty( MAVEN_USER_HOME_PROPERTY_KEY );
-        if ( mavenUserHome == null )
-        {
-            mavenUserHome = System.getenv( MAVEN_USER_HOME_ENV_KEY );
-        }
-
-        return mavenUserHome == null ? DEFAULT_MAVEN_USER_HOME : Paths.get( mavenUserHome );
+        return DEFAULT_MAVEN_USER_HOME;
     }
 }


jorsol:
Sorry, but I don't understand why it has to be a fixed location.

This is actually a useful feature and I fail see why remove it.

gzm55:
wrapper should better share the same behavior with maven, avoid unexpected misunderstanding on a maven scope location "maven user home". if this base dir is wrapper specific, its names should be something like "wrapper user home".

maven original choice of fixed ~/.m2 may not be very convenient, but easy to understand and integrate with other tools.

gzm55:
`mavenUserHome` is more like a maven directory for [user configuration](https://github.com/apache/maven/blob/maven-3.8.5/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java#L133), and expected to work as maven. if we need a location to store maven distribution, it should be named with the purpose, such like `maven wrapper home`, or `maven distribution directory`.

jorsol:
> mavenUserHome is more like a maven directory for [user configuration](https://github.com/apache/maven/blob/maven-3.8.5/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java#L133), and expected to work as maven.

It's not only for user configuration, it's also used for `maven.repo.local` (the repository on the local machine), that BTW can also be configured. By removing this you are making it fixed into `~/.m2` and that already contradicts your perception of user configuration.

The "naming" of `maven.user.home` might or might not be correct, that could be debated, it was borrowed from [gradle](https://docs.gradle.org/current/userguide/directory_layout.html#dir:gradle_user_home) and the previous wrapper was already using it.

I honestly don't mind if it's renamed to something like `maven.distro.basedir`, and it should be properly documented, but just removing this without replacement doesn't sounds like a good plan.

Anyway, the final word should come from a Maven maintainer.

gzm55:
> The "naming" of maven.user.home might or might not be correct, that could be debated, it was borrowed from [gradle](https://docs.gradle.org/current/userguide/directory_layout.html#dir:gradle_user_home) and the previous wrapper was already using it.

yes, that's the gradle pattern, not maven. In maven, some paths are fixed or only configure via cli options, such as ~/.m2/{settings.xml, toolchainx.xml}, $basedir/.mvn/{extensions.xml, maven.config, jvm.config}. whereas `maven.repo.local` is a configurable location, normally set via ~/.m2/settings.xml or the well-defined property. That is, the configuration locations are almost fixed, and data storage are configurable. To follow this convention, how about provide a property `maven.distro.local`, default to `${user.home}/.m2/wrapper/dists`?

gzm55:
function `mavenUserHome()` now has two usages:

1. in MavenWrapperMain.java, `private static void addSystemProperties( Path rootDir )`. in the function wrapper loads two property files into system. the files are {maven.user.home}/maven.properties and {basedir}/maven.properties. This logic would come from a [document](https://github.com/apache/maven/blob/master/maven-core/src/site/apt/configuration-management.apt) for maven develop team, and maven 2.x, 3.x, 4.x do not load these two path. To avoid `mvn ...` and `mvnw ...` running with different system properties, `addSystemProperties()` should not be called within wrapper.
2. pass `mavenUserHome()` to `PathAssembler.mavenUserHome` which will only affect `PathAssembler.getBaseDir()`. (PS: `getBaseDir()` return working dir when user pass "PROJECT" to `getBaseDir()`, it should be a bug). This base dir is used as a storage base for distribution and unzip files.

Then i would prefer add a property `maven.wrapper.repo.local` default to ~/.m2.

jira-importer:
Resolve #195

jira-importer:
Resolve #195

PR #30, opened 2022-03-13
[MWRAPPER-59] Inline MavenWrapperDownloader.java into mvnw

gzm55:
By inlining a lean version of MavenWrapperDownloader.java into `mnvw` script, we can

  - remove the `source` installing type
  - avoid reuse an outdated MavenWrapperDownloader.class
  - make the downloading part of `mvnw` simpler


slawekjaranowski:
You removed test `maven-wrapper-plugin/src/it/projects/type_source/`
How inline downloader are covered by tests?

gzm55:
> You removed test `maven-wrapper-plugin/src/it/projects/type_source/` How inline downloader are covered by tests?

the IT for type_source verify that the _plugin_ can download/extract the correct files, then run `./mvnw -v` can:

- produce some logs
- download the maven-wrapper.jar

The verify process do not mock the environment without neither `curl` nor `wget`, so the java source branch is not covered.

The inlined Downloader is very simple than the previous one, and should less likely go wrong. And to test the function of  the inlined java source in IT stage, we need hide curl and wget commands before run mvnw; or test in the UT stage, by greping out the java source from the script and run a download task.

gzm55:
@slawekjaranowski to save reviewers time, i will close this pr before #33 is decided

jira-importer:
Resolve #219

PR #33, opened 2022-03-16
[MWRAPPER-61] Try to skip maven-wrapper.jar, call mvn directly from mvnw.

gzm55:

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MWRAPPER) filed 
       pulling in other changes.
 - [x] Format the pull request title like `[MWRAPPER-XXX] - Fixes bug in ApproximateQuantiles`,
       commit message.

 - [x] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

The most behavior of new implementation is the same as original.

**For Posix shell `mvnw`:**

1.  dists path pattern is shorten to `~/.m2/wrapper/dists/apache-maven-<version>/<java-string-hash>`
2. follow posix shell, test on macos, centos 6/7/8, alpine with bash 3/4/5, zsh, dash
3. fix all shellcheck.net warnings
4. inline java source of the fallback way for downloading dist zip fille
5. improve password security, avoid leaking to command line or trace log
6. support busybox wget to download maven dist
7. dependency:
  - a posix shell at `/bin/sh`
  - posix built-in commands such a `[`, `echo` and `printf`
  - directory operation commands: `mktemp`, `mkdir`, `mv`, `rm`
  - `cygpath` if Cygwin or MinGW
  - one of three: `wget`, `curl` or `jdk 1.7+`
  - one of two: `unzip` or `tar + gzip`

**For windows cmd batch `mvnw.cmd`:**

1. dists path pattern is shorten to `~/.m2/wrapper/dists/apache-maven-<version>/<md5-hash>`
2. the main functions of start, download and install parts are implemented via powershell, which is preinstalled since Win7 SP1
3. fix warnings given by powershell PSScriptAnalyzer
4. dependency:
  - `cmd`
  - `powershell 5.1+`

### DIFF
--- a/maven-wrapper-distribution/pom.xml
+++ b/maven-wrapper-distribution/pom.xml
@@ -57,6 +57,7 @@ under the License.
         </executions>
         <configuration>
           <descriptors>
+            <descriptor>src/assembly/only-script.xml</descriptor>
             <descriptor>src/assembly/script.xml</descriptor>
             <descriptor>src/assembly/source.xml</descriptor>
             <descriptor>src/assembly/bin.xml</descriptor>
--- a/maven-wrapper-distribution/src/assembly/only-script.xml
+++ b/maven-wrapper-distribution/src/assembly/only-script.xml
@@ -1,0 +1,63 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<assembly xmlns="http://maven.apache.org/ASSEMBLY/2.1.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/ASSEMBLY/2.1.0 http://maven.apache.org/xsd/assembly-2.1.0.xsd">
+  <id>only-script</id>
+  <formats>
+    <format>zip</format>
+  </formats>
+  <includeBaseDirectory>false</includeBaseDirectory>
+  <files>
+    <file>
+      <source>src/resources/only-mvnw.cmd</source>
+      <destName>mvnw.cmd</destName>
+      <outputDirectory>.</outputDirectory>
+      <lineEnding>dos</lineEnding>
+      <fileMode>0644</fileMode>
+      <filtered>true</filtered>
+    </file>
+    <file>
+      <source>src/resources/only-mvnwDebug.cmd</source>
+      <destName>mvnwDebug.cmd</destName>
+      <outputDirectory>.</outputDirectory>
+      <lineEnding>dos</lineEnding>
+      <fileMode>0644</fileMode>
+      <filtered>true</filtered>
+    </file>
+    <file>
+      <source>src/resources/only-mvnw</source>
+      <destName>mvnw</destName>
+      <outputDirectory>.</outputDirectory>
+      <lineEnding>unix</lineEnding>
+      <fileMode>0755</fileMode>
+      <filtered>true</filtered>
+    </file>
+    <file>
+      <source>src/resources/only-mvnwDebug</source>
+      <destName>mvnwDebug</destName>
+      <outputDirectory>.</outputDirectory>
+      <lineEnding>unix</lineEnding>
+      <fileMode>0755</fileMode>
+      <filtered>true</filtered>
+    </file>
+  </files>
+</assembly>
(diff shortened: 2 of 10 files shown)


maven-wrapper-distribution/src/resources/mvnw
```diff
@@ -19,305 +19,160 @@
...
+	{
+	  protected java.net.PasswordAuthentication getPasswordAuthentication()
+	  {
+	    return new java.net.PasswordAuthentication( System.getenv( "MVNW_USERNAME" ), System.getenv( "MVNW_PASSWORD" ).toCharArray() );
+	  }
+	  public static void main( String[] args ) throws Exception
+	  {
+	    setDefault( new Downloader() );
+	    java.nio.file.Files.copy( new java.net.URL( args[0] ).openStream(), java.nio.file.Paths.get( args[1] ).toAbsolutePath().normalize() );
+	  }
```
slawekjaranowski, line 161: I saw it in another PR ... please don't mix.

jorsol:
While trying to skip the maven-wrapper.jar might be a nice goal, this creates a dependency on some external tools like `unzip`, and on some environments like containers this is an issue:

```bash
docker run -it --rm eclipse-temurin:17.0.2_8-jdk sh -c unzip
sh: 1: unzip: not found
```

So, -1 from my side.

Also, opening two different PRs (https://github.com/apache/maven-wrapper/pull/35) that are related to the same issue and that both touch the same files is not a good approach.

gzm55:
> While trying to skip the maven-wrapper.jar might be a nice goal, this creates a dependency on some external tools like `unzip`, and on some environments like containers this is an issue:
> 
> ```shell
> docker run -it --rm eclipse-temurin:17.0.2_8-jdk sh -c unzip
> sh: 1: unzip: not found
> ```
> 
> So, -1 from my side.
> 
> Also, opening two different PRs (#35) that are related to the same issue and that both touch the same files is not a good approach.

ok, #35 is closed.

thanks to point out the unzip dependency. i think this dependency is not a big problem. on this environment, we can still depends on the java source.

gzm55:
@jorsol new commit works in docker image eclipse-temurin:17.0.2_8-jdk, now decompression depends on either unzip or gzip

gzm55:
@slawekjaranowski can we allow [shellcheck action](https://github.com/marketplace/actions/shellcheck)?

hboutemy:
@gzm55 what is required to do that?

FYI, I'll do the 3.1.1 release soon (without this change), then we'll work on merging this PR that fully changes the approach: we'll need to define if it's best to call it 3.2.0 or even 4.0.0...

gzm55:
@hboutemy 

3.2.0 or  4.0.0 are both OK.
Although implementation changes, the cli interface and the dependencies are almost remain same, so, imo, a minor version bump is enough.

hboutemy:
the bump in version is not because of visible change, but a so vast reimplementation that there is a de-facto a risk that someone finds an edge case

gzm55:
how about these steps for accepting this:

gzm55:
> the bump in version is not because of visible change, but a so vast reimplementation that there is a de-facto a risk that someone finds an edge case


@hboutemy how about these steps for accepting this:

1. review to find most hiding risks, if _accept_ this implementation, then
2. publish as a new `type`, such as `only-script` for one release
3. replace `script` type with `only-script`, also keep the original script as `script0`, for one or two release
4. remove other types at the last stage

slawekjaranowski:
my proposition in order to go faster with it 
1. in this PR
 - create new `type` - `only-script`
 - add IT test
 - rebase to one commit
 - add text to PR description ... `I hereby declare this contribution to be licenced under the...` see .github/pull_request_template.md
 
2. next PR 
 - extends - or create documentations about provided type with requirements
 
3. release next version ... get feedbacks

gzm55:
- [ ] create new type - only-script
- [ ] add IT test
- [ ] rebase to one commit
- [ ] add text to PR description ... I hereby declare this contribution to be licenced under the..

gzm55:
@slawekjaranowski @hboutemy 

A new type `only-script` is added to keep the compatibility of all the previous types. Some basic document for the new type is also described in `index.apt.vm`

maven-wrapper/src/main/java/org/apache/maven/wrapper/Installer.java
```diff
@@ -70,8 +70,12 @@ public Path createDist( WrapperConfiguration configuration )
...
             Logger.info( "Detected MVNW_REPOURL environment variable " + mvnwRepoUrl );
             String mvnPath = distributionUrl.toURL().toString();
-            mvnPath = mvnPath.substring( mvnPath.indexOf( "org/apache/maven" ) );
-            distributionUrl = new URI( mvnwRepoUrl ).resolve( "/" ).resolve( mvnPath );
+            int pathIdx =  mvnPath.indexOf( "org/apache/maven" );
+            if ( 0 <= pathIdx )
+            {
+                mvnPath = mvnPath.substring( pathIdx );
+                distributionUrl = new URI( mvnwRepoUrl ).resolve( "/" ).resolve( mvnPath );
+            }
```
slawekjaranowski, line 78: How it is connected with new script type?

`pathIdx` will be `-1` if there is no substring ....

gzm55:
@slawekjaranowski the `Installer.java` is rollback, cause the issue discussed within this java source has already been fixed via [`MWRAPPER-68`](https://github.com/apache/maven-wrapper/commit/d9f0b1db825b5facdc65adf1211e9a4a5e472c29).

gzm55:
@hboutemy @slawekjaranowski any updates?

slawekjaranowski:
@gzm55  big thanks

jira-importer:
Resolve #221

jira-importer:
Resolve #221

PR #35, opened 2022-03-19
[MWRAPPER-61] Try to skip maven-wrapper.jar, call mvn.cmd directly from mvnw.cmd

gzm55:
This is the cmd batch part of #33 .

1. most behavior is the same as maven-wrapper.
2. dists path pattern is shorten to `~/.m2/wrapper/dists/apache-maven-<version>/<hash>`,
   using powershell default file hash
3. the main functions of  start, download and install parts are implemented via powershell, which is preinstalled
   since Win7 SP1


jira-importer:
Resolve #221

jira-importer:
Resolve #221

PR #38, opened 2022-04-07
(doc) Fixing doc about installing mvnDebug-scripts

speedlog:
Adaptation of documentation to current behavior: https://github.com/apache/maven-wrapper/blob/master/maven-wrapper-plugin/src/main/java/org/apache/maven/plugins/wrapper/WrapperMojo.java#L98


slawekjaranowski:
Thanks.

PR #44, opened 2022-05-17
[MWRAPPER-67] Remove '\r' when using IFS on windows for wrapperUrl from mvnw script usage such as git bash

hazendaz:
Usage of git bash will not download the maven wrapper jar along with curl (probably others) due to having windows line endings in the URL (trailing).  To ensure that is not the case, make sure to strip invalid line endings out before usage.

Use case, ./mvnw in powershell will use mvnw.cmd and has no issues downloading.  If user does same in git bash, it will fail with invalid URL error with curl.  Using ./mvnw.cmd there will work but not natural usage.  To ensure this simply just works for full support, trim out invalid line feeds.

note: This only affected the download.  It worked otherwise.

 - [ x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MWRAPPER) filed 
       pulling in other changes.
 - [ x] Each commit in the pull request should have a meaningful subject line and body.
 - [ x] Format the pull request title like `[MWRAPPER-XXX] - Fixes bug in ApproximateQuantiles`,
       commit message.
 - [ x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
 - [x ] Run `mvn clean verify` to make sure basic checks pass. A more thorough check will 
 - [x ] You have run the integration tests successfully (`mvn -Prun-its clean verify`).

 - [ x] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

### DIFF
--- a/maven-wrapper-distribution/src/resources/mvnw
+++ b/maven-wrapper-distribution/src/resources/mvnw
@@ -199,7 +199,9 @@ else
       wrapperUrl="https://repo.maven.apache.org/maven2/org/apache/maven/wrapper/maven-wrapper/@@project.version@@/maven-wrapper-@@project.version@@.jar"
     fi
     while IFS="=" read -r key value; do
-      case "$key" in (wrapperUrl) wrapperUrl="$value"; break ;;
+      # Remove '\r' from value to allow usage on windows as IFS does not consider '\r' as a separator ( considers space, tab, new line ('\n'), and custom '=' )
+      safeValue=$(echo "$value" | tr -d '\r')
+      case "$key" in (wrapperUrl) wrapperUrl="$safeValue"; break ;;
       esac
     done < "$MAVEN_PROJECTBASEDIR/.mvn/wrapper/maven-wrapper.properties"
     log "Downloading from: $wrapperUrl"


hazendaz:
The script file can be either windows or Unix based as that doesn't seem to cause issue.  Possibly it's from the read of wrapper properties.  Tried printing the data there but all looked ok.  Adding this did fix it.

Sent from my Verizon, Samsung Galaxy smartphone
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

michael-o:
I Prefer the source to be fixed.

hazendaz:
@michael-o The issue is from how the property file is read.  Its reading the entire line so it gets line feed.  It really should not pick up the line feed.  I have changed the logic to fix at source of issue at least removing exact same as I did initially.  So basically the same fix just at root of the issue.  The file itself doesn't need fixed.  Its irrelevent.  The problem is the read from that file picking up unnecessary content.  Making the file either windows or unix based is IMHO not the right fix.

maven-wrapper-distribution/src/resources/mvnw
```diff
@@ -192,7 +192,8 @@ else
       wrapperUrl="https://repo.maven.apache.org/maven2/org/apache/maven/wrapper/maven-wrapper/@@project.version@@/maven-wrapper-@@project.version@@.jar"
     fi
     while IFS="=" read key value; do
-      case "$key" in (wrapperUrl) wrapperUrl="$value"; break ;;
+      # Remove invalid line endings from value
+      case "$key" in (wrapperUrl) wrapperUrl="${value%$'\r'}"; break ;;
```
michael-o, line 196: ...and you are sure this runs on a POSIX only shell?

hazendaz: My testing was on Windows 10, Windows 2016 server, and RHEL 8.  So I don't know for sure but the logic there is posix compliant per multiple sources such as [here](https://www.ibm.com/docs/en/aix/7.1?topic=shell-parameter-substitution-in-korn-posix) and [here](https://unix.stackexchange.com/questions/389209/what-kinds-of-string-interpolation-does-posix-sh-support).

hazendaz: @michael-o When you get a chance, can you check in on this again?  Thanks.

michael-o: Not this month. I have basically no interest in the wrapper since I consider it more or less pointless.

hazendaz:
@michael-o Ping, any possibility of this being merged before next release happens?

hazendaz:
rebased, adjusted title, squashed commits.

hazendaz:
Adjusted to be posix compliant.  This will strip out any \r but if \r\n at end of line, it strips both but that should be fine given this is used for property value pairs and not expecting \n either but IFS is ignoring that one by design.

Sample test I did on this.  In sample I have mixed in \r\n between words and at end to better show removal of \r along with fact last \n will get removed due to placement.

```
$ value=$'firstLine\r\nsecondLine\r\n'

$ set | grep "^value="
value=$'firstLine\r\nsecondLine\r\n'

$ safeValue=$(echo "$value" | sed 's/\r$//')

$ set | grep "^safeValue="
safeValue=$'firstLine\nsecondLine'
```

bmarwell:
So, can this be reproduced by using GitHub actions and a bash shell?

hazendaz:
I think this repo would need gitattributes added to have the issue.  Not sure what git does when that is missing and not telling it how to handle line endings.

basically this would need present

```
* text=auto
```

So that windows run is checked out and files end up with \r\n in them...

ascopes:
This issue also affects another area, which I have put a PR up for at #86.

maven-wrapper-distribution/src/resources/mvnw
```diff
@@ -194,7 +194,9 @@ else
       wrapperUrl="https://repo.maven.apache.org/maven2/org/apache/maven/wrapper/maven-wrapper/@@project.version@@/maven-wrapper-@@project.version@@.jar"
     fi
     while IFS="=" read -r key value; do
-      case "$key" in (wrapperUrl) wrapperUrl="$value"; break ;;
+      # Remove '\r' from value to allow usage on windows as IFS does not consider '\r' as a separator ( considers space, tab, new line ('\n'), and custom '=' )
+      safeValue=$(echo "$value" | sed 's/\r$//')
```
ascopes, line 198: using

```
`command`
```

may be favourable compared to

```
$(...)
```

...as some sources specify that $() doesn't appear to be supported by csh.

maven-wrapper-distribution/src/resources/mvnw
```diff
@@ -194,7 +194,9 @@ else
       wrapperUrl="https://repo.maven.apache.org/maven2/org/apache/maven/wrapper/maven-wrapper/@@project.version@@/maven-wrapper-@@project.version@@.jar"
     fi
     while IFS="=" read -r key value; do
```
ascopes, line 196: as a side note, don't think it is necesarilly related to this issue, but https://docs.oracle.com/cd/E23095_01/Platform.93/ATGProgGuide/html/s0204propertiesfileformat01.html suggests that `:` is a valid character here too.

maven-wrapper-distribution/src/resources/mvnw
bmarwell: -1. See: https://www.shellcheck.net/wiki/SC2006
We use `/bin/sh` which  should never point too `csh`, only  to posix compliant shells (bash, dash on Ubuntu).
If you do this, shellcheck will fail.

hazendaz:
any status here?

maven-wrapper-distribution/src/resources/mvnw
slawekjaranowski: Can you use `tr` instead of `sed` - it should be enough

bmarwell: Or just use a bash variable expansion so we don't need to spawn another process.
Given everyone is on bash 4 now.

michael-o:
@bmarwell If you think this one is good, please go ahead and merge.

maven-wrapper-distribution/src/resources/mvnw
norrisjeremy: > Or just use a bash variable expansion so we don't need to spawn another process. Given everyone is on bash 4 now.

That would be wrong assumption.
On macOS:
```
$ /bin/sh --version
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin22)
Copyright (C) 2007 Free Software Foundation, Inc.
```

michael-o: > Or just use a bash variable expansion so we don't need to spawn another process. Given everyone is on bash 4 now.

NEVER expect something to be true in a target environment. That's Windows point of view.

ascopes: > Or just use a bash variable expansion

Didnt you say this was using /bin/sh? Surely that means there is no guarantee you have bash in the environment at all?

hazendaz: correct, my original code had bash expansion but that fails posix.  Suggested change from 'sed' to 'tr' though makes sense.  Looking at that now.

hazendaz:
Changed 'sed' to 'tr' per @slawekjaranowski request.  Rebased and minor text changes on the commit message.

jira-importer:
Resolve #265

jira-importer:
Resolve #265

PR #65, opened 2022-10-11
[MWRAPPER-80] support maven-mvnd

gzm55:
Try to integrate maven-mvnd as a maven distribution. The `mvnd` use a native executable as entry or a new entry class in generic version to communicate with the background daemon. So we can only execute the `mvnd` by the wrapper script with `only-script` distribution type. Now `only-mvnw` (for *nix OS) and `only-mvnw.cmd` (for windows) work fine without sha256 checksum verification.

- manual select platform: distributionUrl=https://downloads.apache.org/maven/mvnd/0.8.2/maven-mvnd-0.8.2-darwin-amd64.zip
- auto select platform: distributionUrl=https://downloads.apache.org/maven/mvnd/0.8.2/maven-mvnd-0.8.2-bin.zip

In the auto mode, the script selects a proper platform via the result of "uname -a", or fallback to the pure java version for unknown platform. For `distributionSha256Sum`, we also need new pattern to support multiple checksums, such as `distributionSha256Sum=darwin-aarch64:111,linux-amd64:222`

 - [X] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MWRAPPER) filed 
       pulling in other changes.
 - [X] Format the pull request title like `[MWRAPPER-XXX] - Fixes bug in ApproximateQuantiles`,
       commit message.

### DIFF
--- a/maven-wrapper-distribution/src/resources/only-mvnw
+++ b/maven-wrapper-distribution/src/resources/only-mvnw
@@ -98,21 +98,20 @@ done < "${0%/*}/.mvn/wrapper/maven-wrapper.properties"
 
 case "${distributionUrl##*/}" in
 (maven-mvnd-*bin.*)
-  _MVNW_USE_MVND=true MVN_CMD=mvnd _MVNW_REPO_PATTERN=/maven/mvnd/
+  MVN_CMD=mvnd.sh _MVNW_REPO_PATTERN=/maven/mvnd/
   case "${PROCESSOR_ARCHITECTURE-}${PROCESSOR_ARCHITEW6432-}:$(uname -a)" in
-  (*AMD64:CYGWIN*|*AMD64:MINGW*) distributionPlatform=windows-amd64; MVN_CMD=mvnd.exe ;;
+  (*AMD64:CYGWIN*|*AMD64:MINGW*) distributionPlatform=windows-amd64 ;;
   (:Darwin*x86_64) distributionPlatform=darwin-amd64 ;;
   (:Darwin*arm64) distributionPlatform=darwin-aarch64 ;;
   (:Linux*x86_64*) distributionPlatform=linux-amd64 ;;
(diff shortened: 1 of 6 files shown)


nicoloboschi (approved):

nicoloboschi:
This is a great feature!
@bmarwell @amitmisra-ww @slawekjaranowski PTAL

slawekjaranowski:
please also add IT test - to see how it works

What with additional task per OS, like for macOs `xattr -r -d ...`?

https://github.com/apache/maven-mvnd#install-manually

gzm55:
> What with additional task per OS, like for macOs `xattr -r -d ...`?
> 
> https://github.com/apache/maven-mvnd#install-manually

The "com.apple.quarantine" extended attribute is set when the zip/tgz files are downloaded by the browsers, and not set when downloaded by curl/wget/java. This is tested on macos since 10.15. And the original maven distribution installed by wrapper also relies on this behavior. So the `xattr -r -d ...` step is not required by the wrapper for both original maven and mvnd.

bmarwell:
> and not set when downloaded by curl/wget/java

Yes this is true, for whatever reason.

gzm55:
> please also add IT test - to see how it works

IT test is added. The test case success on linux and macos, but failed on windows since not implemented in `mvnw.cmd`.

gzm55:
@slawekjaranowski @bmarwell `mvnw` and `mvnw.cmd` both pass the IT test.

gzm55:
@slawekjaranowski could you give some suggestions or review for this pr?

slawekjaranowski:
If we support only `only-script` WrapperMojo should detect it and set proper `distributionType` in other case it will not works ... I'm right?

gzm55:
> If we support only `only-script` WrapperMojo should detect it and set proper `distributionType` in other case it will not works ... I'm right?

yes, a fast-fail check is added in `execute()` of the mojo

jira-importer:
Resolve #270

jira-importer:
Resolve #270

PR #117, opened 2024-02-01
[MWRAPPER-123]  hash string char-by-char for AIX's ksh implementation

bmarwell:
AIX's ksh's `printf %d` implementation cannot handle full string input.

Thus, create the hash char-by-char. This is what the implementation would have done anyway.

---

 - [X] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MWRAPPER) filed 
       pulling in other changes.
 - [X] Format the pull request title like `[MWRAPPER-123] - Fixes bug in ApproximateQuantiles`,
       commit message.

### DIFF
--- a/maven-wrapper-distribution/src/resources/only-mvnw
+++ b/maven-wrapper-distribution/src/resources/only-mvnw
@@ -82,7 +82,8 @@ set_java_home() {
 hash_string() {
   str="${1:-}" h=0
   while [ -n "$str" ]; do
-    h=$(((h * 31 + $(LC_CTYPE=C printf %d "'$str")) % 4294967296))
+    char="${str%"${str#?}"}"
+    h=$(((h * 31 + $(LC_CTYPE=C printf %d "'$char")) % 4294967296))
     str="${str#?}"
   done
   printf %x\\n $h


maven-wrapper-distribution/src/resources/only-mvnw
```diff
@@ -72,7 +72,9 @@ set_java_home() {
 hash_string() {
   str="${1:-}" h=0
   while [ -n "$str" ]; do
-    h=$(( ( h * 31 + $(LC_CTYPE=C printf %d "'$str") ) % 4294967296 ))
+    # TODO: convert the input 'str' to an int for hash calculation
+    # printf '%d' will only convert the first character, and breaks on some ksh implementations.
+    h=$(( ( h * 31 + 104 ) % 4294967296 ))
```
rmannibucau, line 77: dont think we can do that due to next line which mutate str so str is not a constant in the loop, it shifts the first ascii char one by one so I guess a better fix is to use the substitution so do a substring of 1 char to let all shell handle the "string to char" (`'`) conversion working

maven-wrapper-distribution/src/resources/only-mvnw
```diff
@@ -70,11 +70,19 @@ set_java_home() {
...
-    str="${str#?}"
+  str="${1}"
+  length=${#str}
+  h=0
+  # strings start with index 1
+  idx=1
+
+  # lt: skip line break at end
+  while [ $idx -lt "$length" ]; do
+    char=$( printf "%s" "$str" | cut -c $idx )
```
bmarwell, line 81: Fork here, alternative is to use string index (supported by ksh and bash and zsh). However:

```
In maven-wrapper-distribution/src/resources/only-mvnw line 80:
    char=${str:${idx}:1}
         ^-------------^ SC3057 (warning): In POSIX sh, string indexing is undefined.
```

bmarwell: The other advantage is that string index starts at 0 😄

bmarwell: According to Ubuntu, they still use Dash as default symlink target of `/bin/sh` which does not know about variable substitution: https://wiki.ubuntu.com/DashAsBinSh

They recommend using awk (but cut is enough here)

rmannibucau (approved):
ok for me but kind of yet another proof we don't want to have a shell script with any logic so I would move to a java/jsh or anything else script since wrapper will no more be used with older jvm version

bmarwell:
> ok for me but kind of yet another proof we don't want to have a shell script with any logic so I would move to a java/jsh or anything else script since wrapper will no more be used with older jvm version

100 %

I still don't like the subshell (`| cut `). If anyone can come up with a better solution, let us know.

slawekjaranowski:
My test:

```
hash_string() {
  str="${1:-}" h=0
  while [ -n "$str" ]; do
    c="${str%${str#?}}"
    h=$(( ( h * 31 + $(LC_CTYPE=C printf %d "'$c") ) % 4294967296 ))
    str="${str#?}"
  done
  printf %x\\n $h
}
```

@bmarwell can you try on AIX?

bmarwell:
> My test:
> 
> ```
> hash_string() {
>   str="${1:-}" h=0
>   while [ -n "$str" ]; do
>     c="${str%${str#?}}"
>     h=$(( ( h * 31 + $(LC_CTYPE=C printf %d "'$c") ) % 4294967296 ))
>     str="${str#?}"
>   done
>   printf %x\\n $h
> }
> ```
> 
> @bmarwell can you try on AIX?

sadly...

```
# vi test.sh
# ksh test.sh "https://reopsitory/repository/maven-central/org/apache/maven/apache-maven/3.9.6/apache-maven-3.9.6-bin.zip"
printf: 2589392689: A return value of a math subroutine is not within machine precision.
7fffffff
```

slawekjaranowski:
just out of curiosity, what result will be:

```
echo 'T="abcde"; echo "${T#?}"' | bash  

echo 'T="abcde"; echo "${T%${T#?}}"' | bash
```

bmarwell:
```
# echo 'T="abcde"; echo "${T#?}"' | bash
bcde

# echo 'T="abcde"; echo "${T%${T#?}}"' | bash
a
```

slawekjaranowski:
> ```
> # echo 'T="abcde"; echo "${T#?}"' | bash
> bcde
> 
> # echo 'T="abcde"; echo "${T%${T#?}}"' | bash
> a
> ```

I thought about test on AIX with ksh ...

bmarwell:
> > ```
> > # echo 'T="abcde"; echo "${T#?}"' | bash
> > bcde
> > 
> > # echo 'T="abcde"; echo "${T%${T#?}}"' | bash
> > a
> > ```
> 
> I thought about test on AIX with ksh ...

```
# echo 'T="abcde"; echo "${T#?}"' | ksh
bcde

echo 'T="abcde"; echo "${T%${T#?}}"' | ksh
a
```

I think I know what you have in mind...

bmarwell:
I think we can replace this

```
  # strings start with index 1
  idx=1

  # lt: skip line break at end
  while [ $idx -lt "$length" ]; do
    char=$( printf "%s" "$str" | cut -c $idx )
    h=$(( ( h * 31 + $(LC_CTYPE=C printf %d "'$char") ) % 4294967296 ))
    idx=$(( idx + 1 ))
  done
``` 

with this:

```
  while [ "$str" != "" ]; do
    char="${str%${str#?}}"
    h=$(( ( h * 31 + $(LC_CTYPE=C printf %d "'$char") ) % 4294967296 ))
    str="${T#?}"
  done
```

slawekjaranowski (approved):

jira-importer:
Resolve #245

PR #127, opened 2024-04-17
[MWRAPPER-125] Make only-script the default

cstamas:
This makes only-script the default.

---

https://issues.apache.org/jira/browse/MWRAPPER-125


slawekjaranowski (approved):

jira-importer:
Resolve #293

PR #128, opened 2024-04-17
[MWRAPPER-131] Make downloader not stream into final target file

cstamas:
Stream into temp file in same dir, and at the end move the stuff into file place.

---

https://issues.apache.org/jira/browse/MWRAPPER-131


cstamas:
MavenWrapperDownloader formatting is off (is not palantir), so will leave it as is for review, and before merge reformat

jira-importer:
Resolve #269

PR #129, opened 2024-04-17
[MWRAPPER-130] Update parent POM to 42

cstamas:
Meaning of life.

---

https://issues.apache.org/jira/browse/MWRAPPER-130


slawekjaranowski (approved):
We can remove junit-bom from depManagemnt - it is in parent

cstamas:
Removed

jira-importer:
Resolve #312

PR #131, opened 2024-04-17
[MWRAPPER-103] Workaround: demote check to Warning

cstamas:
For start, instead to die off, let's demote the failure to warning level.

---

https://issues.apache.org/jira/browse/MWRAPPER-103

### DIFF
--- a/maven-wrapper-distribution/src/resources/only-mvnw.cmd
+++ b/maven-wrapper-distribution/src/resources/only-mvnw.cmd
@@ -124,7 +124,7 @@ if ($distributionSha256Sum) {
     Write-Error "Checksum validation is not supported for maven-mvnd. `nPlease disable validation by removing 'distributionSha256Sum' from your maven-wrapper.properties."
   }
   if ((Get-FileHash "$TMP_DOWNLOAD_DIR/$distributionUrlName" -Algorithm SHA256).Hash.ToLower() -ne $distributionSha256Sum) {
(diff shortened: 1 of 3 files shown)


cstamas:
@gsmet eyes please. Disclaimer: I am not Win guy, so if you can bring on someone to verify, please do so.

cstamas:
@jorsol eyes please

jorsol:
The issue affects the `wrapperSha256Sum` from `mvnw`, not `only-mvnw`, actually the only-mvnw does not validate the `wrapperSha256Sum` since it doesn't download the jar.

In the end, this is not the fix for the issue reported, the trouble is that `only-mvnw` does different things than `mvnw`, so it seems that using `only-mvnw` works and `mvnw` don't.

cstamas:
Oh, got it now. Ok, am closing this one out then, as I missed the topic.

jorsol:
Since the default is now `only-mvnw`, maybe the current workaround is to simply remove the lines from:

https://github.com/apache/maven-wrapper/blob/c9a3e67e5295c41b4a9bc4fd4babe81d20e7e18f/maven-wrapper-distribution/src/resources/mvnw.cmd#L156-L172

Or comment the  `if ERRORLEVEL 1 goto error`.

cstamas:
Yup, that is my problem as well: "not sure" :smile:

cstamas:
Next attempt: https://github.com/apache/maven-wrapper/pull/132

jira-importer:
Resolve #292

jira-importer:
Resolve #292

PR #132, opened 2024-04-17
[MWRAPPER-103] Apply recommended workaround

cstamas:
As Marcono1234 on JIRA issue recommended, added workaround just before the use of `Get-FileHash` call.

---

https://issues.apache.org/jira/browse/MWRAPPER-103

### DIFF
--- a/maven-wrapper-distribution/src/resources/only-mvnw.cmd
+++ b/maven-wrapper-distribution/src/resources/only-mvnw.cmd
@@ -123,6 +123,7 @@ if ($distributionSha256Sum) {
   if ($USE_MVND) {
     Write-Error "Checksum validation is not supported for maven-mvnd. `nPlease disable validation by removing 'distributionSha256Sum' from your maven-wrapper.properties."
   }
+  Import-Module $PSHOME\Modules\Microsoft.PowerShell.Utility -Function Get-FileHash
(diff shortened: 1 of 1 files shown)


maven-wrapper-distribution/src/resources/mvnw.cmd
```diff
@@ -160,6 +160,7 @@ FOR /F "usebackq tokens=1,2 delims==" %%A IN ("%MAVEN_PROJECTBASEDIR%\.mvn\wrapp
 )
 IF NOT %WRAPPER_SHA_256_SUM%=="" (
     powershell -Command "&{"^
+       "Import-Module $PSHOME\Modules\Microsoft.PowerShell.Utility -Function Get-FileHash"^
```
jorsol, line 163: I assume that a `;` is needed at the end of Get-FileHash.

cstamas: Hm, no idea, but Google found me this: https://poshcode.gitbook.io/powershell-practice-and-style/style-guide/code-layout-and-formatting#avoid-using-semicolons-as-line-terminators

cstamas: even if it is to avoid, will add to be consistent with existing PS code....

michael-o (approved):

jira-importer:
Resolve #292

PR #139, opened 2024-05-12
MWRAPPER-139: Remove trailing whitespace from fields in properties

ascopes:
MWRAPPER-139: Remove trailing whitespace from fields in maven-wrapper.properties

This prevents issues due to confusion between CRLF and LF when running in more exotic locations such as MINGW on Windows but using ./mvnw rather than ./mvnw.cmd. A prime example of where this is an issue would be on windows-latest GitHub Actions runners.

Fixes MWRAPPER-139, MWRAPPER-137 (which I didn't spot for some reason when reviewing the
existing tickets on JIRA... might have been a blip in JIRA that didn't display it?).

### DIFF
--- a/maven-wrapper-distribution/src/resources/only-mvnw
+++ b/maven-wrapper-distribution/src/resources/only-mvnw
@@ -97,11 +97,19 @@ die() {
   exit 1
 }
 
+trim() {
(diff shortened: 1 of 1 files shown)


jira-importer:
Resolve #279

PR #142, opened 2024-05-21
[MWRAPPER-140] only-script type does not honor MAVEN_USER_HOME

cstamas:
User provided patch to "only-script" type (as bin does support it).

---

https://issues.apache.org/jira/browse/MWRAPPER-140


jira-importer:
Resolve #315

jira-importer:
Resolve #315

PR #155, opened 2024-09-19
[MWRAPPER-150] - Fails to validate checksums on MacOS Sequoia

jon-signal:
The version of `sha256sum` that ships with macOS Sequoia appears to have a bug where it does not read from stdin by default, even though its `man` page says, in part (emphasis added):

> In all cases, each file listed on the command line is processed separately.  **If no files are listed on the command line**, or a file name is given as -, input is taken from stdin instead.

…but invoking `sha256` as it appears in the `mvnw` script does not work as expected:

```sh
❯ echo '3d8f20ce6103913be8b52aef6d994e0c54705fb527324ceb9b835b338739c7a8  /Users/jon/REDACTED/.mvn/wrapper/maven-wrapper.jar' | sha256sum -c
usage: sha256sum [-bctwz] [files ...]
```

…which in turn results in a non-successful exit code, which in turn leads to a (rather alarming!) warning about a mismatched checksums and possible compromise.

Adding a `-` explicitly calls for `sha256sum` to read from stdin.

----

### DIFF
--- a/maven-wrapper-distribution/src/resources/mvnw
+++ b/maven-wrapper-distribution/src/resources/mvnw
@@ -284,7 +284,7 @@ done <"$MAVEN_PROJECTBASEDIR/.mvn/wrapper/maven-wrapper.properties"
 if [ -n "$wrapperSha256Sum" ]; then
   wrapperSha256Result=false
   if command -v sha256sum >/dev/null; then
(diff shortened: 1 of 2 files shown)


slachiewicz (approved):

jira-importer:
Resolve #297

cstamas:
@slawekjaranowski ping

PR #158, opened 2024-10-30
[MWRAPPER-111] Trim whitespace when reading from properties file

mhalbritter:
JIRA issue: https://issues.apache.org/jira/projects/MWRAPPER/issues/MWRAPPER-111

This PR trims the values read from the property files by removing whitespace such as newlines.

This has already be done in the case of `maven-wrapper-distribution/src/resources/only-mvnw`, but it's missing from `maven-wrapper-distribution/src/resources/mvnw`.

With this change, validating the wrapper checksum no longer fails if the property file has CLRF line endings.


slachiewicz (approved):

mhalbritter:
Thanks @slachiewicz !

jira-importer:
Resolve #307

jira-importer:
Resolve #307

PR #337, opened 2025-07-16
Add JDK management with Foojay Disco API integration

gnodet:
## 🚀 JDK Management for Maven Wrapper

This PR adds comprehensive JDK management capabilities to Maven Wrapper using the **Foojay Disco API**, enabling automatic JDK download and installation for the `only-script` distribution type.

### 🎯 Automatic JDK Management
- **Automatic JDK download and installation** for `only-script` distribution
- **34+ JDK distributions** supported via Foojay Disco API
- **Version resolution**: Major versions (e.g., `17`) resolve to latest patch versions
- **Multi-platform support**: Linux, macOS, Windows with architecture detection
- **Toolchain integration**: Automatic `toolchains.xml` generation for multi-JDK builds
- **Dynamic LTS detection**: Uses Disco API to identify LTS versions (no hardcoded heuristics)
- **Bypass option**: `MVNW_SKIP_JDK` environment variable to use system JDK

### 🔧 Configuration Options
```properties
# Basic JDK configuration
jdkVersion=17
jdkDistribution=temurin

# Toolchain JDK for multi-version builds
toolchainJdkVersion=11
toolchainJdkDistribution=corretto

# Update policies
jdkUpdatePolicy=daily  # never, daily, weekly, monthly, always, interval:X

# Direct URLs (optional)
jdkDistributionUrl=https://example.com/jdk.tar.gz
jdkSha256Sum=abc123...
```

### 🆔 Environment Variables
```bash
# Skip JDK management and use system JDK
export MVNW_SKIP_JDK=true

# Enable verbose output
export MVNW_VERBOSE=true

# Configure JDK settings
export MAVEN_WRAPPER_JDK_VERSION=17
export MAVEN_WRAPPER_JDK_DISTRIBUTION=temurin
```

### 📦 Supported Distributions

**Popular Distributions:**
- `temurin` - Eclipse Adoptium (default)
- `corretto` - Amazon Corretto
- `zulu` - Azul Zulu
- `liberica` - BellSoft Liberica
- `oracle_open_jdk` - Oracle OpenJDK
- `microsoft` - Microsoft OpenJDK
- `semeru` - IBM Semeru

**Specialized Distributions:**
- `graalvm_ce11`, `graalvm_ce17` - GraalVM Community Edition
- `sap_machine` - SAP Machine
- `dragonwell` - Alibaba Dragonwell
- `jetbrains` - JetBrains Runtime
- `bisheng` - Huawei BiSheng
- `kona` - Tencent Kona
- `mandrel` - Red Hat Mandrel

**Complete list**: [Foojay Disco API](https://api.foojay.io/disco/v3.0/distributions)

### Maven Plugin
```bash
# Generate wrapper with JDK management
mvn wrapper:wrapper -Dtype=only-script -Djdk=17 -DjdkDistribution=temurin

# With toolchain support
mvn wrapper:wrapper -Dtype=only-script -Djdk=21 -DjdkDistribution=corretto -DtoolchainJdk=17 -DtoolchainJdkDistribution=temurin
```

### Properties Configuration
```properties
# Production: Pin exact version
jdkVersion=17.0.14
jdkDistribution=temurin
jdkUpdatePolicy=never

# Development: Auto-update to latest patches
jdkVersion=17
jdkDistribution=temurin
jdkUpdatePolicy=daily
```

### Bypassing JDK Management
```bash
# Use system JDK instead of wrapper-managed JDK
export MVNW_SKIP_JDK=true
./mvnw clean compile

# Windows
set MVNW_SKIP_JDK=true
mvnw.cmd clean compile

# Useful for CI matrix testing
MVNW_SKIP_JDK=true ./mvnw test
```

### Multi-JDK Toolchain Setup
```properties
# Main JDK for build
jdkVersion=21
jdkDistribution=temurin

# Toolchain JDK for compilation
toolchainJdkVersion=17
toolchainJdkDistribution=corretto
```

### Architecture
- **Shell-based implementation** for `only-script` distribution type
- **No Java dependency** for JDK installation (avoids chicken-and-egg problem)
- **Foojay Disco API integration** for JDK resolution and download
- **Cross-platform support** with native shell scripts (Unix) and PowerShell (Windows)
- **API URL constants** for maintainability and easy version updates

### Dynamic LTS Detection
- **Real-time LTS detection** via Disco API `/major_versions` endpoint
- **Intelligent caching** to avoid repeated API calls
- **Robust fallback** to hardcoded LTS list if API unavailable
- **Accurate warnings** showing current LTS versions (6, 7, 8, 11, 17, 21)
- **Future-proof** - automatically includes new LTS versions when released

### Distribution Validation
- **Comprehensive validation** against known Disco API distributions
- **Helpful error messages** showing all available distributions
- **Typo prevention** with clear distribution suggestions

### Bypass Mechanisms
- **MVNW_SKIP_JDK**: Skip JDK management entirely, use system JDK
- **Direct URLs**: Override version/distribution resolution with exact URLs
- **Environment variables**: Configure JDK settings without modifying properties

### Caching & Performance
- **Intelligent caching** with configurable update policies
- **SHA-256 verification** for security
- **Efficient downloads** with resume support
- **Version resolution caching** to minimize API calls

### Security Features
- **HTTPS downloads** by default
- **SHA-256 checksum verification** for all downloads
- **Automatic checksum resolution** from Disco API
- **Manual checksum override** for direct URLs

### For Developers
- **Zero JDK setup** - automatic download and installation
- **Consistent environments** across team members
- **Easy distribution switching** for testing
- **Toolchain support** for multi-JDK projects
- **Accurate LTS guidance** with real-time data
- **Flexible bypass options** for special cases

### For CI/CD
- **Reproducible builds** with exact version pinning
- **No pre-installed JDK requirement** in build environments
- **Flexible update policies** for different environments
- **Multi-platform consistency**
- **Matrix testing support** with MVNW_SKIP_JDK

### For Projects
- **Simplified onboarding** - no JDK installation instructions needed
- **Version consistency** across all environments
- **Easy JDK upgrades** via configuration changes
- **Professional distribution support** (34+ options)
- **Troubleshooting options** with bypass mechanisms

## 🔄 Migration Path

Existing projects can easily adopt JDK management:

1. **Switch to `only-script`** distribution type
2. **Add JDK configuration** to `maven-wrapper.properties`
3. **Remove manual JDK setup** from documentation
4. **Enjoy automatic JDK management**

```bash
# Simple migration command
mvn wrapper:wrapper -Dtype=only-script -Djdk=17 -DjdkDistribution=temurin
```

## 🧪 Testing

- ✅ **Cross-platform testing** (Linux, macOS, Windows)
- ✅ **Multiple JDK distributions** verified
- ✅ **Version resolution** tested (major → patch versions)
- ✅ **Toolchain integration** validated
- ✅ **Error handling** and validation tested
- ✅ **Caching and update policies** verified
- ✅ **Dynamic LTS detection** confirmed with API
- ✅ **MVNW_SKIP_JDK bypass** functionality tested

## 📚 Documentation

Comprehensive documentation added:
- **JDK_MANAGEMENT.md** - Complete feature guide
- **Configuration examples** for all use cases
- **Environment variable documentation**
- **Troubleshooting guide** for common issues
- **Migration instructions** for existing projects
- **MVNW_SKIP_JDK usage examples**

## 🎉 Summary

This PR transforms Maven Wrapper into a **complete JDK management solution** that:
- **Eliminates JDK setup friction** for developers
- **Ensures consistent environments** across teams
- **Supports 34+ professional JDK distributions**
- **Provides flexible configuration options** for all use cases
- **Maintains security** with checksum verification
- **Offers excellent performance** with intelligent caching
- **Uses dynamic LTS detection** with real-time API data
- **Requires zero maintenance** for future JDK versions
- **Includes bypass mechanisms** for special cases and troubleshooting

The implementation is **production-ready**, **well-tested**, and **thoroughly documented**, making it easy for projects to adopt automatic JDK management while maintaining flexibility for edge cases.

### DIFF
--- a/maven-wrapper-plugin/src/main/java/org/apache/maven/plugins/wrapper/WrapperMojo.java
+++ b/maven-wrapper-plugin/src/main/java/org/apache/maven/plugins/wrapper/WrapperMojo.java
@@ -20,16 +20,24 @@ package org.apache.maven.plugins.wrapper;
 
 import javax.inject.Inject;
 
+import java.io.BufferedReader;
 import java.io.BufferedWriter;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.net.HttpURLConnection;
+import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Properties;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.apache.maven.Maven;
 import org.apache.maven.execution.MavenSession;
@@ -61,6 +69,9 @@ public class WrapperMojo extends AbstractMojo {
     private static final String MVNW_REPOURL = "MVNW_REPOURL";
 
     protected static final String DEFAULT_REPOURL = "https://repo.maven.apache.org/maven2";
+
+    // Disco API constants
+    private static final String DISCO_API_BASE_URL = "https://api.foojay.io/disco/v3.0";
 
     // CONFIGURATION PARAMETERS
 
@@ -161,6 +172,77 @@ public class WrapperMojo extends AbstractMojo {
      */
     @Parameter(property = "distributionUrl")
     private String distributionUrl;
+
+    /**
+     * The JDK version to use for Maven execution.
+     * Can be any valid JDK release, such as "17", "21", or "17.0.14".
+     */
+    @Parameter(property = "jdk")
+    private String jdkVersion;
+
+    /**
+     * The distribution of JDK to download using native Disco API names.
+     * Supported distributions: temurin, corretto, zulu, liberica, oracle_open_jdk, microsoft, semeru, graalvm_ce11, etc.
+     * Default is temurin.
+     */
+    @Parameter(property = "jdkDistribution", defaultValue = "temurin")
+    private String jdkDistribution;
+
+    /**
+     * Direct URL for main JDK distribution download.
+     * If specified, overrides jdkVersion and jdkDistribution.
+     */
+    @Parameter(property = "jdkUrl")
+    private String jdkDistributionUrl;
+
+    /**
+     * The JDK version to use for toolchains.
+     * Can be any valid JDK release.
+     */
+    @Parameter(property = "toolchainJdk")
+    private String toolchainJdkVersion;
+
+    /**
+     * The distribution of JDK to download for toolchains using native Disco API names.
+     * Supported distributions: temurin, corretto, zulu, liberica, oracle_open_jdk, microsoft, semeru, graalvm_ce11, etc.
+     * Default is temurin.
+     */
+    @Parameter(property = "toolchainJdkDistribution", defaultValue = "temurin")
+    private String toolchainJdkDistribution;
+
+    /**
+     * Direct URL for toolchain JDK distribution download.
+     * If specified, overrides toolchainJdkVersion and toolchainJdkDistribution.
+     */
+    @Parameter(property = "toolchainJdkUrl")
+    private String toolchainJdkDistributionUrl;
+
+    /**
+     * SHA-256 checksum for the main JDK distribution
+     */
+    @Parameter(property = "jdkSha256Sum")
+    private String jdkSha256Sum;
+
+    /**
+     * SHA-256 checksum for the toolchain JDK distribution
+     */
+    @Parameter(property = "toolchainJdkSha256Sum")
+    private String toolchainJdkSha256Sum;
+
+    /**
+     * JDK update policy for major version resolution. Controls how often the latest patch version is checked.
+     * Supported values: never, daily, always, interval:X (where X is minutes).
+     * Default is daily.
+     */
+    @Parameter(property = "jdkUpdatePolicy", defaultValue = "daily")
+    private String jdkUpdatePolicy;
+
+    /**
+     * Skip JDK version stability warnings for non-LTS versions or specific version pinning.
+     * Default is false (warnings are shown).
+     */
+    @Parameter(property = "skipJdkWarnings", defaultValue = "false")
+    private boolean skipJdkWarnings;
 
     // READONLY PARAMETERS
 
@@ -340,10 +422,42 @@ public class WrapperMojo extends AbstractMojo {
                 out.append("wrapperSha256Sum=" + wrapperSha256Sum + System.lineSeparator());
             }
             if (alwaysDownload) {
-                out.append("alwaysDownload=" + Boolean.TRUE + System.lineSeparator());
+                out.append("alwaysDownload=true" + System.lineSeparator());
             }
             if (alwaysUnpack) {
-                out.append("alwaysUnpack=" + Boolean.TRUE + System.lineSeparator());
+                out.append("alwaysUnpack=true" + System.lineSeparator());
+            }
+            if (jdkVersion != null) {
+                out.append("jdkVersion=" + jdkVersion + System.lineSeparator());
+            }
+            if (jdkDistribution != null) {
+                out.append("jdkDistribution=" + jdkDistribution + System.lineSeparator());
+            }
+            if (jdkDistributionUrl != null) {
+                out.append("jdkDistributionUrl=" + jdkDistributionUrl + System.lineSeparator());
+            }
+            if (toolchainJdkVersion != null) {
+                out.append("toolchainJdkVersion=" + toolchainJdkVersion + System.lineSeparator());
+            }
+            if (toolchainJdkDistribution != null) {
+                out.append("toolchainJdkDistribution=" + toolchainJdkDistribution + System.lineSeparator());
+            }
+            if (toolchainJdkDistributionUrl != null) {
+                out.append("toolchainJdkDistributionUrl=" + toolchainJdkDistributionUrl + System.lineSeparator());
+            }
+            if (jdkSha256Sum != null) {
+                out.append("jdkSha256Sum=" + jdkSha256Sum + System.lineSeparator());
+            }
+            if (toolchainJdkSha256Sum != null) {
+                out.append("toolchainJdkSha256Sum=" + toolchainJdkSha256Sum + System.lineSeparator());
+            }
+            if (jdkUpdatePolicy != null) {
+                out.append("jdkUpdatePolicy=" + jdkUpdatePolicy + System.lineSeparator());
+            }
+
+            // Show JDK version stability warnings at installation time
+            if (!skipJdkWarnings) {
+                showJdkVersionWarnings();
             }
         } catch (IOException ioe) {
             throw new MojoExecutionException("Can't create maven-wrapper.properties", ioe);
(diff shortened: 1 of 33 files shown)


mosabua:
Whoa .. thats amazing!

gnodet:
> Is my understanding correct that the `mvnw` script now downloads and installs a JDK distrubution instead of using the system JDK?

That's not completely exact.  The JDK download is an opt-in if you add the required properties.  The default behaviour is unchanged.  There's also a `MVNW_SKIP_JDK` env var to skip using the JDK indicated by those properties (can be useful when running in a JDK matrix in a CI job).

src/site/markdown/JDK_MANAGEMENT.md
```diff
@@ -0,0 +1,597 @@
...
+
+## Supported JDK Distributions
+
+The Maven Wrapper supports 34+ JDK distributions through the Foojay Disco API using native distribution names:
+
+### Popular Distributions
+
+| Distribution | Description | Recommended Use |
+|-------------|-------------|-----------------|
+| `temurin` | Eclipse Adoptium (recommended) | General purpose, excellent support |
```
norrisjeremy, line 135: Is it appropriate to recommend a specific distribution?

gnodet: Definitely no, I have switched that to `default` rather than `recommended`, as we need one.  But it came back ;-)

norrisjeremy: FYI, the PR description also states Temurin is recommended.
Dunno if you want to update that too?

norrisjeremy:
> > To confirm my understanding: if we add the new JDK properties into `.mvn/wrapper/maven-wrapper.properties`, but then set the `MVNW_SKIP_JDK=true` env variable, `mvnw` will use the system JDK as always and not attempt to download a JDK?
> 
> Correct

Perfect, thanks! This sounds like a great feature which we can use with our dev team to allow easy bootstrapping of the preferred JDK distribution on their workstations, whilst continuing to have our Github Actions CI pipelines use the system provided JDK.

maven-wrapper-plugin/src/main/java/org/apache/maven/plugins/wrapper/WrapperMojo.java
```diff
@@ -405,4 +519,184 @@ protected String determineRepoUrl(String envRepoUrl, Settings settings) {
...
+
+    private boolean checkJdkVersionStability(String jdkType, String version, String distribution) {
+        if (version == null) {
+            return false;
+        }
+
+        boolean warningShown = false;
+
+        // Check for non-LTS versions
+        if (isNonLtsVersion(version)) {
```
Bukama, line 555: I assume you also want this for Maven 3? Otherwise maybe use multiline string

src/site/markdown/JDK_MANAGEMENT.md
```diff
@@ -0,0 +1,597 @@
...
+## Configuration
+
+### Basic JDK Configuration
+
+Add JDK configuration to your `.mvn/wrapper/maven-wrapper.properties` file:
+
+```properties
+# JDK Management
+jdkVersion=17                # Resolves to latest 17.x (e.g., 17.0.14)
+jdkDistribution=temurin      # Disco API distribution name
```
Bukama, line 54: Disco API was never mentioned before, not suddenly pops up in a comment. Should be mentioned before so people can look up whats available

src/site/markdown/JDK_MANAGEMENT.md
```diff
@@ -0,0 +1,597 @@
...
+# Amazon Corretto
+jdkVersion=17          # → 17.0.15-amzn
+jdkDistribution=corretto
+```
+
+**Note**: Major version resolution (e.g., `17` → `17.0.14+7`) is recommended as it automatically provides the latest security patches and bug fixes.
+
+## Supported JDK Distributions
+
+The Maven Wrapper supports 34+ JDK distributions through the Foojay Disco API using native distribution names:
```
Bukama, line 129: I would remove the number and only mentioned Disco API. Also add link to there

Bukama (approved):

gzm55:
Exciting feature!

@gnodet 
is it possible to support the `.java-version` file for defining the default JDK version of a project? It is normally generated by jenv to lock a project local JDK.

gzm55:
> # Useful for CI matrix testing
> MVNW_SKIP_JDK=true ./mvnw test

@gnodet how about to support selecting dynamicly JDK version via an environment, then the mvnw can pin almost the jdk matrix by project.

maven-wrapper-plugin/src/main/java/org/apache/maven/plugins/wrapper/WrapperMojo.java
gnodet: Not sure what you mean here ?

Bukama: You log a lot of single line strings to generate something that looks like a multiline string? Since java 13 (?) Java supports texts blocks like. As we set Java 17 for Maven 4, maybe we can make use of this here if the change is only aiming at Maven 4.

gnodet: Ah, got it.  So this is currently not focused on Maven 4. Also, using multiline strings changes the output, as only the first line will be prepended with `[WARN] ` afaik, so it kinda messes the output which is not correctly aligned anymore.

gnodet:
> > `#` Useful for CI matrix testing
> > MVNW_SKIP_JDK=true ./mvnw test
> 
> @gnodet how about to support selecting dynamicly JDK version via an environment, then the mvnw can pin almost the jdk matrix by project.

I've added env vars to override the JDK properties.

gzm55:
> > > `#` Useful for CI matrix testing
> > > MVNW_SKIP_JDK=true ./mvnw test
> > 
> > 
> > @gnodet how about to support selecting dynamicly JDK version via an environment, then the mvnw can pin almost the jdk matrix by project.
> 
> I've added env vars to override the JDK properties.

should we make all the env vars all has the same prefix `MVNW_`

carlspring:
I'm not on the committers list, but this looks like fantastic new feature! :)

gnodet:
> > > > `#` Useful for CI matrix testing
> > > > MVNW_SKIP_JDK=true ./mvnw test
> > > 
> > > 
> > > @gnodet how about to support selecting dynamicly JDK version via an environment, then the mvnw can pin almost the jdk matrix by project.
> > 
> > 
> > I've added env vars to override the JDK properties.
> 
> should we make all the env vars all has the same prefix `MVNW_`

They do AFAIK, did I miss something ?

gzm55:
> > > > > `#` Useful for CI matrix testing
> > > > > MVNW_SKIP_JDK=true ./mvnw test
> > > > 
> > > > 
> > > > @gnodet how about to support selecting dynamicly JDK version via an environment, then the mvnw can pin almost the jdk matrix by project.
> > > 
> > > 
> > > I've added env vars to override the JDK properties.
> > 
> > 
> > should we make all the env vars all has the same prefix `MVNW_`
> 
> They do AFAIK, did I miss something ?

`maven-wrapper-plugin/src/it/projects/jdk_environment_variables/pom.xml` left some `MAVEN_WRAPPER_`

olamy:
It's a great feature.
Maybe we shouldn't download some binaries per default but ask the user before downloading the idk?

gnodet:
> It's a great feature. Maybe we shouldn't download some binaries per default but ask the user before downloading the idk?

Not sure to understand.  The default behaviour is unchanged, in order to have the JDK downloaded, the user needs to opt-in and add the needed properties, which will trigger the JDK download, same as the wrapper downloads the Maven distribution.  This can be opt-out using env-var too.  You would like an additional confirmation before actually downloading the JDK or the maven distribution ?

olamy:
> > It's a great feature. Maybe we shouldn't download some binaries per default but ask the user before downloading the idk?
> 
> Not sure to understand. The default behaviour is unchanged, in order to have the JDK downloaded, the user needs to opt-in and add the needed properties, which will trigger the JDK download, same as the wrapper downloads the Maven distribution. This can be opt-out using env-var too. You would like an additional confirmation before actually downloading the JDK or the maven distribution ?

I don't mind for Maven because it's Maven :) but for jdk is a different (we do not maintain and we don't even know what is the license) maybe it would be better to have additional confirmation (can be disable eventually by the generator of the script) 

I just tried this:
```
mvn org.apache.maven.plugins:maven-wrapper-plugin:3.3.3-SNAPSHOT:wrapper -Dtype=only-script -Djdk=17 -DjdkDistribution=temurin 
```
Now if I push the generated script any user of the opensource project where I did that will have automatic download of a jdk binaries.

jimisola:
Great feature! From a user perspective (especially if I wear the Enterprise and not the Open Source Developer one) I have few things:

* Do you plan to release this feature a new major version of Maven Wrapper? I consider this a major change  (although not necessarily breaking) since the default behaviour changes and the user has to opt-out using `MVNW_SKIP_JDK`. That is, a simple update of Maven Wrapper should need cause me to suddently have a bunch of JDK installed without my knowledge (especially, since we use SDKMan already).
* Is there a way that this could be integrated with SDKMan for users that want that? 
* Have you considered security concerns with the distribution url? A person might check the source code, the POM etc for security stuff but fail to understand the implication of a distibution url that leads to a malicious JDK? I reckon if you use  jdkDistribution then it uses the Foojay Disco API which seems to be some sort of assurance of the integrity of that JDK.

norrisjeremy:
> * Have you considered security concerns with the distribution url? A person might check the source code, the POM etc for security stuff but fail to understand the implication of a distibution url that leads to a malicious JDK? I reckon if you use  jdkDistribution then it uses the Foojay Disco API which seems to be some sort of assurance of the integrity of that JDK.

Isn't the fact that you can pin the SHA-256 hash via the `jdkSha256Sum` property cover the security perspectives?

gnodet:
> Great feature! From a user perspective (especially if I wear the Enterprise and not the Open Source Developer one) I have few things:
> 
> * Do you plan to release this feature a new major version of Maven Wrapper? I consider this a major change  (although not necessarily breaking) since the default behaviour changes and the user has to opt-out using `MVNW_SKIP_JDK`. That is, a simple update of Maven Wrapper should need cause me to suddently have a bunch of JDK installed without my knowledge (especially, since we use SDKMan already).

Not really because the default behaviour DOES NOT change.  You have to first opt-in and specify the JDK version.  Once you have opted-in, you can then opt-out with `MVNW_SKIP_JDK`.  

> * Is there a way that this could be integrated with SDKMan for users that want that?

Not sure.  That was really my first attempt. Unfortunately, sdkman does not keep track of old releases, which is a problem from a reproducibility pov.

> * Have you considered security concerns with the distribution url? A person might check the source code, the POM etc for security stuff but fail to understand the implication of a distibution url that leads to a malicious JDK? I reckon if you use  jdkDistribution then it uses the Foojay Disco API which seems to be some sort of assurance of the integrity of that JDK.

I agree with Jeremy here.  A simple to secure things is to use an exact url + sha256.

If you want to keep some versatility, another way would be to be able to provide an alternative REST endpoint that could be implemented within an entreprise. At runtime, only `https://api.foojay.io/disco/v3.0/packages` is actually used.  So an env var could specify a REST endpoint different than the default `https://api.foojay.io/disco/v3.0`.

jimisola:
> I sort of surprised that an enterprise that is that particular about security would use maven-wrapper in the first place, but to each their own. 🤷‍♂️

It's not a matter of Enterprise or not. When I develop on my spare time I don't necessarily want to have a bunch of JDKs downloaded to my computer without my knowledge.

jimisola:
> > * Do you plan to release this feature a new major version of Maven Wrapper? I consider this a major change  (although not necessarily breaking) since the default behaviour changes and the user has to opt-out using `MVNW_SKIP_JDK`. That is, a simple update of Maven Wrapper should need cause me to suddently have a bunch of JDK installed without my knowledge (especially, since we use SDKMan already).
> 
> Not really because the default behaviour DOES NOT change. You have to first opt-in and specify the JDK version. Once you have opted-in, you can then opt-out with `MVNW_SKIP_JDK`.

Maybe I'm missing something here. Who is that has to opt-in?

If I download an open source project that uses Maven Wrapper and then run it with `mvnw` do I have to opt-in for the download of a JDK (`jdkDistribution`/`jdkDistributionUrl`) to be enabled?
 
> > * Is there a way that this could be integrated with SDKMan for users that want that?
> 
> Not sure. That was really my first attempt. Unfortunately, sdkman does not keep track of old releases, which is a problem from a reproducibility pov.

Agreed. Yes, it is.

> > * Have you considered security concerns with the distribution url? A person might check the source code, the POM etc for security stuff but fail to understand the implication of a distibution url that leads to a malicious JDK? I reckon if you use  jdkDistribution then it uses the Foojay Disco API which seems to be some sort of assurance of the integrity of that JDK.
> 
> I agree with Jeremy here. A simple to secure things is to use an exact url + sha256.

See my use-case above. I don't see how an exact url and an sha256 in an open source project's `maven-wrapper.properties` helps.
 
> If you want to keep some versatility, another way would be to be able to provide an alternative REST endpoint that could be implemented within an entreprise. At runtime, only `https://api.foojay.io/disco/v3.0/packages` is actually used. So an env var could specify a REST endpoint different than the default `https://api.foojay.io/disco/v3.0`.

Yes, I saw that that was added. But, I don't think that it's relevant to the security use-case I wrote above - unless I miss understood something.

gnodet:
> > I sort of surprised that an enterprise that is that particular about security would use maven-wrapper in the first place, but to each their own. 🤷‍♂️
> 
> It's not a matter of Enterprise or not. When I develop on my spare time I don't necessarily want to have a bunch of JDKs downloaded to my computer without my knowledge.

Then you can easily opt-out by default by adding `export MVNW_SKIP_JDK=true` in your shell config.

rmannibucau (approved):

bmarwell:
I would like to test it on AIX (bash and maybe the not so old ksh). The default ksh is buggy...

However, most AIX systems don't have a direct internet connection. Does anyone have access to such a system?

Bukama:
> At least for specific environments (eg CI) off by default? Corporate environments...

yes. I'm on your side - not allowed and/or able to just download what I want.

gnodet:
> > At least for specific environments (eg CI) off by default? Corporate environments...
> 
> yes. I'm on your side - not allowed and/or able to just download what I want.

That would be opting-out based on an env var.  That's already doable.  Just add `export MVNW_SKIP_JDK=true` in your shell setup and you're done.  Or don't use tool that automate your environment setup. 
How do you control before hand which dependencies are downloaded when you run a maven build ?  Maven does not know by default that you're in a corporate env and that you do not want to go to central by default.  Same here... You need to either disable this feature, or make it point to a different endpoint where you can control the JDK being downloaded.

Bukama:
> How do you control before hand which dependencies are downloaded when you run a maven build ?

At my security has an Nexus IQ Server with whitelist/blacklist activated.

olamy:
TBH, this feature without auto download of the JDK is probably useless :).
People use `mvnw` rather than `mvn` on purpose and by choice. 
If some systems doesn't like it just  add the env var or simply use `mvn`

PR #361, opened 2025-08-26
Normalization: Ensure end of file markers per git standards and tabs as spaces

hazendaz:
While reviewing repo, detected 2 files with tab characters instead of spaces and 20 other files that did not adhere to git expectations of empty line at end of file.  This is a no-opt change that ensures the entire repo follows same standard as all other files complied.


slawekjaranowski:
@hazendaz please look at build result

slawekjaranowski:
so we need to use a tab character with `<<-` or remove all whitespace from here-doc and use `<<`

hazendaz:
I believe its fixed now, I removed the leading space from END.

maven-wrapper-distribution/src/resources/only-mvnw
```diff
@@ -202,19 +202,19 @@ elif set_java_home; then
   javaSource="$TMP_DOWNLOAD_DIR/Downloader.java"
   targetZip="$TMP_DOWNLOAD_DIR/$distributionUrlName"
   cat >"$javaSource" <<-END
```
slawekjaranowski, line 204: can we use `<<END` as we don't have a tabs now?

hazendaz: done and rebased.

PR #362, opened 2025-08-26
Support Maven 4 by setting maven.mainClass to MavenCling

hazendaz:
Maven 4 changed how it works, as a result maven wrapper has been broken for a while.  You can see a working usage of this change here https://github.com/spotbugs/spotbugs-maven-plugin/actions/runs/17228298032.  If unable to see, see the github job here that customized updating wrapper to run on maven 4 here https://github.com/spotbugs/spotbugs-maven-plugin/blob/master/.github/workflows/it-maven-4.0.0.yaml by skipping the update to maven 4 and adjusting the wrapper to simply run maven 4 which is addressed by the changes here.

### DIFF
--- a/maven-wrapper-distribution/src/resources/mvnw
+++ b/maven-wrapper-distribution/src/resources/mvnw
@@ -327,6 +327,7 @@ fi
 MAVEN_CMD_LINE_ARGS="$MAVEN_CONFIG $*"
 export MAVEN_CMD_LINE_ARGS
 
+# Maven main class is here to fix maven 4.0.0-beta-5 through 4.0.0-rc-4
 MAVEN_MAIN_CLASS=org.apache.maven.cling.MavenCling
 WRAPPER_LAUNCHER=org.apache.maven.wrapper.MavenWrapperMain
 
--- a/maven-wrapper-distribution/src/resources/mvnw
+++ b/maven-wrapper-distribution/src/resources/mvnw
@@ -327,6 +327,7 @@ fi
 MAVEN_CMD_LINE_ARGS="$MAVEN_CONFIG $*"
 export MAVEN_CMD_LINE_ARGS
 
+MAVEN_MAIN_CLASS=org.apache.maven.cling.MavenCling
 WRAPPER_LAUNCHER=org.apache.maven.wrapper.MavenWrapperMain
 
 # shellcheck disable=SC2086 # safe args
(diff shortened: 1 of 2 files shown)


kwin:
This is no longer necessary with the upcoming Maven 4 version: https://github.com/apache/maven/commit/f77fe3cbb84eba2fcfe77fd6cd1a8d42d7e38408

slawekjaranowski:
And it is deprecated with `only-script` which is default should be works

hazendaz:
Hi @kwin, I'm glad to see maven core is fixed.  However, is there harm in having this as it does nothing and even works still with maven 3?  That at least allows the broken releases to proceed to working state.  I have to keep using this patch for anything I'm testing with maven 4 currently as that fix is not released.  At work I already scaled that out and protect on top so its not a concern for me there but could see people trying to use various beta / rc versions of maven to check for other issues and they will run into this problem.  Maybe a comment added that its to fix specifically the few versions affected would be a compromise.  I'm fine either way though as I'm not impacted at this point for my needs.

Affected versions are: beta-5 to rc-4

hazendaz:
> For me, there's another reason to use the bin distribution type: it's more portable, as a JRE is always needed anyway. I also faced some problems with SHA validation when using script-only in the past. If I recall correctly, it used to fail under Git Bash on Windows.

I find we have same issues at work with sha validation in general.  Another point is that maven didn't actually flag this deprecated.  Saying its deprecated in pull requests or mailing lists doesn't make it so if the code doesn't say so.

ascheman:
I looked into this while triaging M4-related PRs and verified the behavior against the actual `m2.conf` files:

* Maven 3.9.x: `main is org.apache.maven.cli.MavenCli` — hardcoded, `${maven.mainClass}` is never interpolated, so this change is indeed a no-op for Maven 3.
* 4.0.0-beta-5…rc-4: `main is ${maven.mainClass}` **without** a default — broken when launched via wrapper (which boots classworlds directly, bypassing `bin/mvn`); this PR fixes exactly those versions.
* 4.0.0-rc-5+: `set maven.mainClass default org.apache.maven.cling.MavenCling` (f77fe3cbb84) — works without this PR.

One concern with merging it as-is: for rc-5+ the script *overrides* the distribution's default with a hardcoded internal class name, baked into every generated `mvnw` for years. If a future Maven version ever renames its main class (updating its own m2.conf default accordingly), wrapper-launched builds would still force the stale name and fail with `ClassNotFoundException` — today's no-op becomes tomorrow's breakage.

If the goal is keeping beta-5…rc-4 usable, an opt-in would be safer: only pass `-Dmaven.mainClass` when the user explicitly sets e.g. `MVNW_MAIN_CLASS`, instead of always. That preserves the escape hatch for the broken pre-releases without pinning an internal Maven class name in perpetuity.

PR #363, opened 2025-08-26
Ensure Path Traversal is fully addressed in MavenWrapperDownloader

hazendaz:
The original sanitizing patch is not enough to prevent path traversal.  This approach more accurately deals with the security concern.

### DIFF
--- a/maven-wrapper-distribution/src/resources/mvn/wrapper/MavenWrapperDownloader.java
+++ b/maven-wrapper-distribution/src/resources/mvn/wrapper/MavenWrapperDownloader.java
@@ -45,8 +45,11 @@ public final class MavenWrapperDownloader {
         try {
             log(" - Downloader started");
             final URL wrapperUrl = URI.create(args[0]).toURL();
-            final String jarPath = args[1].replace("..", ""); // Sanitize path
-            final Path wrapperJarPath = Paths.get(jarPath).toAbsolutePath().normalize();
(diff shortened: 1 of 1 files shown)


slawekjaranowski:
@hazendaz can you show a case where old method allow path traversal and new not
we cen extract it to simple method a use it in test

slawekjaranowski:
code is used to download 

```shell
("$JAVA_HOME/bin/java" -cp .mvn/wrapper MavenWrapperDownloader "$wrapperUrl" "$wrapperJarPath") || rm -f "$wrapperJarPath"
```

and a variable `wrapperJarPath` is set in script by

```shell
wrapperJarPath="$MAVEN_PROJECTBASEDIR/.mvn/wrapper/maven-wrapper.jar"
```

it looks as constant, so issue is not critical for me

hazendaz:
Hi @slawekjaranowski I agree its not critical, in fact its silly and only point is to shut up static scanners that trip on this file.  As I have an override on top of maven wrapper, I'm ok that this didn't make the cut.  I'll work on testing aspect so maybe this makes it next time.  Snyk is also complaining about SSRF as well but all attempts we have made to silence that haven't worked out yet.  In either case, any hacker that has access to this would have to also make the wrapper even fall back to this and they would have all the files to change behavior anyways so its not really an issue in that regard other than statically on code it is.

Issue #364, opened 2025-08-31
`wrapperVersion` property gets removed when upgrading to Maven Wrapper 3.3.3


breun:
### Affected version

3.3.3

### Bug description

When I install a newer version of Maven Wrapper, `wrapperVersion` is removed from `.mvn/wrapper/maven-wrapper.properties`.

To reproduce (using Maven 3.9.11 as `mvn`):

```
mkdir reproducer
cd reproducer

# Install Maven Wrapper 3.3.2
mvn wrapper:3.3.2:wrapper -Dmaven=3.9.11

# This show `wrapperVersion=3.3.2`
grep wrapperVersion .mvn/wrapper/maven-wrapper.properties

# Update to Maven Wrapper 3.3.3
./mvnw wrapper:3.3.3:wrapper

# This returns nothing, because wrapperVersion has been removed, which I don't expect to happen
grep wrapperVersion .mvn/wrapper/maven-wrapper.properties
```

_Note: previously this description also mentioned `distributionSha256Sum` getting removed during a Maven Wrapper update. I've split this off to its own issue: https://github.com/apache/maven-wrapper/issues/367_

slawekjaranowski:
in order to preserver `distributionSha256Sum` we should read existing properties

breun:
> done in:
> 
> * [[MWRAPPER-143] - The wrapperVersion property is not used in only-script mode #145](https://github.com/apache/maven-wrapper/pull/145)

For update automation tools need a reliable way to determine the current version of Maven Wrapper. The `wrapperVersion` property was a way, although admittedly flawed, because I encounter a lot of projects with `wrapperVersion` bumped manually, but the `mvnw*` scripts from a much older version, because they were not updated along with the property.

I saw that Renovate plans to determine the Maven Wrapper version based on a regex check on the scripts themselves (https://github.com/renovatebot/renovate/pull/37779), which should be more reliable than reading `wrapperVersion`, as long as the format of that comment line it’s parsed from doesn’t change, so it’s still a fragile method.

Could Maven Wrapper provide its version in a more robust (machine-readable) way?

GreyTeardrop:
Reposting some context: the `wrapperVersion` property was added in scope of [MWRAPPER-134](https://issues.apache.org/jira/browse/MWRAPPER-134) by https://github.com/apache/maven-wrapper/pull/135.

slawekjaranowski:
[...]

PR #145, opened 2024-06-17
[MWRAPPER-143] - The wrapperVersion property is not used in only-script mode

bdemers:
The only-script mode doesn't use the wrapper distribution, this property is now only added to the maven-wrapper.properties file when other types are used.

Following this checklist to help us incorporate your 
contribution quickly and easily:

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MWRAPPER) filed 
       for the change (usually before you start working on it).  Trivial changes like typos do not 
       require a JIRA issue.  Your pull request should address just this issue, without 
       pulling in other changes.
 - [x] Each commit in the pull request should have a meaningful subject line and body.
 - [x] Format the pull request title like `[MWRAPPER-XXX] - Fixes bug in ApproximateQuantiles`,
       where you replace `MWRAPPER-XXX` with the appropriate JIRA issue. Best practice
       is to use the JIRA issue title in the pull request title and in the first line of the 
       commit message.
 - [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
 - [x] Run `mvn clean verify` to make sure basic checks pass. A more thorough check will 
       be performed on your pull request automatically.
 - [x] You have run the integration tests successfully (`mvn -Prun-its clean verify`).

If your pull request is about ~20 lines of code you don't need to sign an
[Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) if you are unsure
please ask on the developers list.

To make clear that you license your contribution under 
the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
you have to acknowledge this by using the following check-box.

 - [ ] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

 - [ ] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

Fixes #291

PR #135, opened 2024-04-22
[MWRAPPER-134] Add wrapperVersion to maven-wrapper.properties

cstamas:
That contains the current version of the deployed Maven Wrapper

---

https://issues.apache.org/jira/browse/MWRAPPER-134

PR #365, opened 2025-09-02
Revert wrapper version removal

breun:
Revert #145 for #364, because `wrapperVersion` is used to communicate the Maven Wrapper version to users and automated update tools (see #135).


maven-wrapper-plugin/src/main/java/org/apache/maven/plugins/wrapper/WrapperMojo.java
```diff
@@ -326,6 +326,7 @@ private void replaceProperties(String wrapperVersion, Path targetFolder) throws
                 + buffer().strong("Maven " + mavenVersion) + " and download from " + repoUrl);
 
         try (BufferedWriter out = Files.newBufferedWriter(wrapperPropertiesFile, StandardCharsets.UTF_8)) {
+            out.append("wrapperVersion=" + wrapperVersion + System.lineSeparator());
```
slawekjaranowski, line 329: line 337 can be removed

Issue #395, opened 2026-01-26
mvnw.cmd crashes with "Cannot index into a null array" on 32-bit System Profile path (Jenkins service)


naotsugu:
### Affected version

3.3.3

## Description

In `mvnw.cmd`, the logic introduced by PR #143 to handle symlinks attempts to access an array index on the `Target` property without checking if it is null.

`mvnw.cmd` fails immediately upon execution with a "Cannot index into a null array" error. This issue specifically occurs when the script runs under the **Local System account** in a **32-bit process** (typical for Jenkins Agents running as a Windows Service), where the default `.m2` home resolves to the System Profile directory.

## Error Log

```Plaintext
icm : Cannot index into a null array
At line:1 char:81
+ ... 'mvnw.cmd'; icm -ScriptBlock ([Scriptblock]::Create((Get-Content -Raw ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [Invoke-Command], RuntimeException
    + FullyQualifiedErrorId : NullArray,Microsoft.PowerShell.Commands.InvokeCommandCommand
```

## Environment

- **OS:** Windows Server 2022    
- **Process Architecture:** 32-bit (x86)    
- **User:** `NT AUTHORITY\SYSTEM`    
- **Target Path:** `C:\WINDOWS\system32\config\systemprofile\.m2`    
- **PowerShell:** 5.1    

## Root Cause Analysis

The script contains the following logic to handle symlinks:

```PowerShell
if ((Get-Item $MAVEN_M2_PATH).Target[0] -eq $null) { ... }
```

We verified the behavior of the `Target` property on this specific environment:

1. **Standard Paths (e.g., Workspace drive):** The `Target` property returns an **empty collection**. Accessing `[0]` returns `$null` safely (in PowerShell 5.1).    
2. **System Profile Path (`.m2`):** When accessing the `.m2` directory in the System Profile from a 32-bit process, the `Target` property returns **strictly `$null`**.    

Since `mvnw.cmd` defaults `MAVEN_M2_PATH` to the user home (System Profile), the script attempts to access index `[0]` on `$null`, causing a fatal `RuntimeException`.

## Verification Code & Result

We ran the following diagnostic inside the failing Jenkins environment, specifically targeting the `.m2` directory:

```PowerShell
$basePath = $env:USERPROFILE
[...]

PR #405, opened 2026-03-08
Fix incorrect path resolution for system profile under WOW64 (x86) redirection

fp024:
When `MAVEN_M2_PATH` points to a Windows system profile path (e.g., `C:\Windows\System32\config\systemprofile\.m2`),
running under an x86 (WOW64) process causes the path to be redirected to `C:\Windows\SysWOW64\config\systemprofile\.m2`.
In this case, `(Get-Item -Path $MAVEN_M2_PATH -Force).Target` returns `$null` instead of an array.

The previous code accessed `.Target[0]` directly, which throws an error when `.Target` is `$null` under an x86 process.

This fix stores `.Target` in a variable first and checks its type before indexing.

* Issue: https://github.com/apache/maven-wrapper/issues/395
* See also: https://github.com/fp024/win-file-link-test/blob/main/README_en.md

---

- [x] Run `mvn verify` to make sure basic checks pass.
- [ ] You have run the integration tests successfully (`mvn -Prun-its verify`).

### DIFF
--- a/maven-wrapper-distribution/src/resources/only-mvnw.cmd
+++ b/maven-wrapper-distribution/src/resources/only-mvnw.cmd
@@ -89,10 +89,11 @@ if (-not (Test-Path -Path $MAVEN_M2_PATH)) {
 }
 
 $MAVEN_WRAPPER_DISTS = $null
-if ((Get-Item -Path $MAVEN_M2_PATH -Force).Target[0] -eq $null) {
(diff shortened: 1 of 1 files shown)


fp024:
Hi,
Verified the behavior before and after the fix by installing Jenkins 2.541.2 on a Windows Server 2022 x64 180-day evaluation edition.

* https://github.com/fp024/win-file-link-test/tree/main/java-project

Please take a look. Thank you. 👍

slachiewicz (approved):

PR #415, opened 2026-04-21
fix(wrapper): preserve MVNW_REPOURL when exec maven in only-mvnw

### LINKED ISSUE #368: only-script for Linux clobbers MVNW_REPOURL
### Affected version

3.3.3

### Bug description

Running the Maven Wrapper plugin through the Maven Wrapper's `only-script` distribution type causes `MVNW_REPOURL` to be discarded, leaving the `distributionUrl` written to `maven-wrapper.properties` at its default value. Consequently, `only-script` for Linux is unable to update itself in any environment that needs `MVNW_REPOURL`.

That is, the hypothetical execution

```sh
$ MVNW_REPOURL=asdfasdf ./mvnw org.apache.maven.plugins:maven-wrapper-plugin:3.3.3:wrapper -Dtype=only-script -Dmaven=3.9.11 -X
```

where `asdfasdf` is a functional mirror should have produced an output similar to

```
...
[DEBUG] Using repo URL from MVNW_REPOURL environment variable.
[DEBUG] Determined repo URL to use as asdfasdf
[INFO] Configuring .mvn/wrapper/maven-wrapper.properties to use Maven 3.9.11 and download from asdfasdf
```

but will end up producing

```
...
[DEBUG] Determined repo URL to use as https://repo.maven.apache.org/maven2
[INFO] Configuring .mvn/wrapper/maven-wrapper.properties to use Maven 3.9.11 and download from https://repo.maven.apache.org/maven2
```

and the result is consistent with the log messages.

This happens because `only-mvnw` explicitly unsets `MVNW_REPOURL`:

https://github.com/apache/maven-wrapper/blob/maven-wrapper-3.3.3/maven-wrapper-distribution/src/resources/only-mvnw#L149

so by the time `WrapperMojo` tries to read the variable it sees `null`:

https://github.com/apache/maven-wrapper/blob/maven-wrapper-3.3.3/maven-wrapper-distribution/src/resources/only-mvnw#L149

As far as I can see, this bug does not exist in `only-mvnw.cmd`.

The issue is exacerbated by the user property `distributionUrl` value representing the "final" URl. That is, it is not simply a domain substitution and therefore cannot take the place of `MVNW_REPOURL`.

armorbreak001:
## Problem

In `only-mvnw` (the Linux-only-script distribution), the `exec_maven()` function unsets `MVNW_REPOURL` before exec-ing the Maven process:

```sh
unset MVNW_VERBOSE MVNW_USERNAME MVNW_PASSWORD MVNW_REPOURL || :
```

This means when the Maven Wrapper plugin runs (e.g., `mvn wrapper:wrapper`), it cannot read `MVNW_REPOURL` from the environment, so it falls back to the default repo URL written in `maven-wrapper.properties`.

**Impact:** Users who rely on `MVNW_REPOURL` to point to a custom mirror (common in corporate/proxied environments) cannot use `only-script` mode to update the wrapper — the custom URL is silently dropped.

The Windows counterpart (`only-mvnw.cmd`) does **not** unset `MVNW_REPOURL` — it only clears `MVNW_USERNAME` and `MVNW_PASSWORD`. This inconsistency confirms the bug.

## Fix

Remove `MVNW_REPOURL` from the `unset` line in `exec_maven()`, matching the behavior of `only-mvnw.cmd`.

## Reproduction

```sh
export MVNW_REPOURL=https://your-custom-mirror/repo
./mvnw org.apache.maven.plugins:maven-wrapper-plugin:3.3.3:wrapper -Dtype=only-script -Dmaven=3.9.11 -X
# Before fix: uses default https://repo.maven.apache.org/maven2
# After fix: correctly uses your-custom-mirror/repo
```

Fixes #368

### DIFF
--- a/maven-wrapper-distribution/src/resources/only-mvnw
+++ b/maven-wrapper-distribution/src/resources/only-mvnw
@@ -146,7 +146,7 @@ MAVEN_USER_HOME="${MAVEN_USER_HOME:-${HOME}/.m2}"
 MAVEN_HOME="${MAVEN_USER_HOME}/wrapper/dists/${distributionUrlNameMain-}/$(hash_string "$distributionUrl")"
 
 exec_maven() {
-  unset MVNW_VERBOSE MVNW_USERNAME MVNW_PASSWORD MVNW_REPOURL || :
+  unset MVNW_VERBOSE MVNW_USERNAME MVNW_PASSWORD || :
   exec "$MAVEN_HOME/bin/$MVN_CMD" "$@" || die "cannot exec $MAVEN_HOME/bin/$MVN_CMD"
 }
 


yeikel:
> The Windows counterpart (`only-mvnw.cmd`) does **not** unset `MVNW_REPOURL` — it only clears `MVNW_USERNAME` and `MVNW_PASSWORD`. This inconsistency confirms the bug.
> ## Fix

I think this is a bug in terms of consistency. My suspicion is that the wrapper is used more often in Unix environments, which might explain why this inconsistency has not surfaced so far. That said, this is only speculation, as I do not have data to support it.

Now, regarding `unset MVNW_VERBOSE MVNW_USERNAME MVNW_PASSWORD MVNW_REPOURL`, my assumption is that it is _intended to prevent sensitive data, such as credentials_, from being passed to the Maven process. That said, I am not fully sure if this is the actual intent or why. 

In terms of your change, it does not seem correct because it introduces another bug: If `MVNW_REPOURL` requires authentication, removing the username and password while keeping the URL would cause Maven to fail in setups that depend on those credentials.

I think that the first step before we decide that this code should be changed is to document _why this is here_

bmarwell:
@yeikel But the inconsistency is the UNSET.

Try running for a private repo:

```bash
$ MVNW_REPOURL=https://your-custom-mirror/repo mvn wrapper:wrapper -Dtype=only-script  -Dmaven=4.0.0-rc-5
# works, MVNW_REPOURL is used in .mvn/wrapper/maven-wrapper.properties
# which is: https://your-custom-mirror/repo/org/apache/maven/apache-maven/4.0.0-rc-5/apache-maven-4.0.0-rc-5-bin.zip

$ MVNW_REPOURL=https://your-custom-mirror/repo ./mvnw wrapper:wrapper -Dtype=only-script  -Dmaven=4.0.0-rc-5
# with current mvnw, MVNW_REPOURL is not used, defaults are used in .mvn/wrapper/maven-wrapper.properties
# which is: distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/4.0.0-rc-5/apache-maven-4.0.0-rc-5-bin.zip
```

I find this difference surprising to say the least.

PR #429, opened 2026-07-08
Add AGENTS.md + SECURITY.md security-model pointer for scanner discoverability

potiuk:
**This is a proposal for the PMC to review — please correct, reject, or discuss as needed.** Nothing here is a requirement; the maintainer is the decision-maker.

This adds a small `AGENTS.md` + `SECURITY.md` so an automated scan agent can mechanically discover the project's security model via the conventional `AGENTS.md → SECURITY.md` chain. It points at the Maven family's umbrella threat model, which the PMC merged in `apache/maven` (`THREAT_MODEL.md`) — no model content is duplicated here, just the pointer.

Context: the ASF Security team is preparing the Maven repositories for an automated agentic security scan we're piloting. Such scans refuse to run if the model isn't discoverable by that path. Discoverability is the one hard gate; everything else is suggestion.

Doc-only; questions / pushback welcome.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,32 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Agent Guide for maven-wrapper
+
+This file is read by automated agents (security scanners, code
+analyzers, AI assistants) operating on this repository.
+
+## Security
+
+Security model: [SECURITY.md](./SECURITY.md)
+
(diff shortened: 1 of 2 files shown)


sepe81 (commented):
One small suggestion: Since AGENTS.md and SECURITY.md are specifically designed to be efficiently parsed by automated agents, they might be better optimized with SPDX identifiers instead of the full license header:

```md
<!-- SPDX-License-Identifier: Apache-2.0 -->

# Agent Guide for maven-wrapper
...
```

potiuk:
Yeah w e do that - especially in bigger AGENTS.md / SECURITY.md files where we "fight" for every byte of context -  and if PMC is good with that and local checks for licences will accept that - I am happy to swap. Let me do it now to see how it might work.

potiuk:
I can also adapt in other places if this is something that will work for Maven in general :D @slachiewicz @slawekjaranowski  ?).

slachiewicz:
That's good idea and worth to propagate for more cases across our files. We need to check if this would pass RAT plugin validations and our spotless plugin

slachiewicz (approved):

slachiewicz:
For this case I would stay as is and then if we would have working solution wr can propagate change internally.
Real saving in our case would be to replace headers in Java files - but that requires some kind of confirmation that we are complainant to ASF licence of we use short version in files. Recently added full License to every repo as dedicated file.

potiuk:
> For this case I would stay as is and then if we would have working solution wr can propagate change internally. Real saving in our case would be to replace headers in Java files - but that requires some kind of confirmation that we are complainant to ASF licence of we use short version in files. Recently added full License to every repo as dedicated file.

I think latest RAT already supports SPDX headers. If you would like to that experiment - I thik good idea is to merge it as is and I can use my agents :) - to implement follow-up PRs to try to swap SPDX's .. Happy to help with that.

Issue #331, opened 2025-06-17
A maven wrapper updater action


MikAoJk:
### New feature, improvement proposal

Hi!
A maven version of this gradle action could be great to have, in order for my repos to get a pull request with the newest maven wrapper verison.

https://github.com/marketplace/actions/update-gradle-wrapper-action

mbien:
not sure why this would need a new action. Dependabot learned to update poms (https://github.com/dependabot/dependabot-core/tree/main/maven), I don't see why it couldn't update the wrapper too.

MikAoJk:
> not sure why this would need a new action. Dependabot learned to update poms (https://github.com/dependabot/dependabot-core/tree/main/maven), I don't see why it couldn't update the wrapper too.

Hmm, im not sure.
But someway of automatic update wrapper would be gold 🥇

Issue #249, opened 2024-04-22
[MWRAPPER-134] Provide a reliable way to determine the Maven Wrapper version


jira-importer:
**[Nils Breunese](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=breun)** opened **[MWRAPPER-134](https://issues.apache.org/jira/browse/MWRAPPER-134?redirect=false)** and commented

I use [Renovate](https://github.com/renovatebot/renovate) to automatically keep dependencies up-to-date and I noticed that the upgrade to Maven Wrapper 3.3.0 removed the `wrapperUrl` line from `{}.mvn/wrapper/maven-wrapper.properties{`}. I guess that is related to MWRAPPER-120.

A couple of my colleagues implemented the support for keeping Maven Wrapper up-to-date in Renovate, and I believe Renovate currently determines the version of Maven Wrapper based on the `wrapperUrl` property, so I think that logic may need to be updated after this change. Without the `wrapperUrl` property, I think currently the only option to determine the Maven Wrapper version is parsing the `mvnw` / `mvnw.cmd` scripts themselves, but that sounds like a clunky and unreliable way.

In a discussion on the maven-users mailinglist, the idea was posted to add a dedicated property for this (e.g. `{}wrapperVersion{`}), which can be read by external tools.

---

**Affects:** 3.3.0

**Issue Links:**
- [MWRAPPER-120](https://issues.apache.org/jira/browse/MWRAPPER-120) Maven wrapper plugin should not create wrapperUrl property for script-only mode

- [MWRAPPER-135](https://issues.apache.org/jira/browse/MWRAPPER-135) Provide a reliable way to determine the Maven Wrapper type

**Remote Links:**
- [GitHub Pull Request #135
](https://github.com/apache/maven-wrapper/pull/135)

jira-importer:
**[Damiano Albani](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=dalbani)** commented

> [Nils Breunese](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=breun) 
> A couple of my colleagues implemented the support for keeping Maven Wrapper up-to-date in Renovate, and I believe Renovate currently determines the version of Maven Wrapper based on the `wrapperUrl` property, so I think that logic may need to be updated after this change.

Do you have information to share about the change in Renovate Bot to support this new "wrapperVersion"?
I couldn't find anything relevant on https://github.com/renovatebot/renovate/.

[...]

PR #436, opened 2026-07-24
Feature/maven version range

giulong:

- [x] Run `mvn verify` to make sure basic checks pass.
- [x] You have run the integration tests successfully (`mvn -Prun-its verify`).

---

# Problem
As per the [usage docs](https://maven.apache.org/tools/wrapper/maven-wrapper-plugin/usage.html), bumping the Maven version requires manually going on [Maven Central](https://central.sonatype.com/artifact/org.apache.maven/apache-maven/versions) to get the latest. It would be nice having the plugin resolve the highest available version automatically.

# Proposed solution
Maven already supports [Dependency Version Requirement Specification](https://maven.apache.org/pom.html#Dependency_Version_Requirement_Specification), allowing the definition of a range instead of a fixed one. Though it's not advisable to rely on such a mechanism to manage regular dependencies of an application for the sake of reproducible builds, the wrapper is an external tool for which this issue doesn't apply.

Available Maven versions are taken from [maven-metadata.xml](https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/maven-metadata.xml).

# Rationale
Having a native way to automatically bump the Maven version let us schedule a simple wrapper execution. For instance:
* `./mvnw wrapper:wrapper -Dmaven=3.9.16`: fixed versions are considered valid ranges, so no breaking change.
* `./mvnw wrapper:wrapper -Dmaven="[3.0,4.0-alpha)"`: allows the resolution of the latest `3.x.x` version, excluding currently non-stable `4.x.x` versions.

Moreover, this is done without relying on external tools such as GitHub actions (as proposed in #331) or bots like Renovate (as proposed in #249).

### DIFF
--- a/maven-wrapper-plugin/pom.xml
+++ b/maven-wrapper-plugin/pom.xml
@@ -125,6 +125,12 @@ under the License.
       <version>${version.mockito}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.codehaus.plexus</groupId>
+      <artifactId>plexus-utils</artifactId>
+      <version>3.6.1</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/maven-wrapper-plugin/src/it/projects/mavenversion_range/invoker.properties
+++ b/maven-wrapper-plugin/src/it/projects/mavenversion_range/invoker.properties
@@ -1,0 +1,20 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+# 
+#   http://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# `maven=[3.0,3.10-rc)` would fail on Windows
+# since the closing round bracket is not escaped
+invoker.os.family = !windows
--- a/maven-wrapper-plugin/src/it/projects/mavenversion_range/pom.xml
+++ b/maven-wrapper-plugin/src/it/projects/mavenversion_range/pom.xml
@@ -1,0 +1,66 @@
+<?xml version='1.0' encoding='UTF-8'?>
+
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>org.apache.maven.plugins.it.wrapper</groupId>
+  <artifactId>mavenversion_range</artifactId>
+  <version>1.0.0-SNAPSHOT</version>
+  <packaging>pom</packaging>
+
+  <properties>
+    <cmd></cmd>
+  </properties>
+
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.codehaus.mojo</groupId>
+          <artifactId>exec-maven-plugin</artifactId>
+          <version>@version.exec-maven-plugin@</version>
+          <configuration>
+            <executable>mvnw${cmd}</executable>
+            <arguments>
+              <argument>-v</argument>
+            </arguments>
+            <environmentVariables>
+              <MVNW_VERBOSE>true</MVNW_VERBOSE>
+            </environmentVariables>
+          </configuration>
+        </plugin>
+      </plugins>
+    </pluginManagement>
+  </build>
+
+  <profiles>
+    <profile>
+      <id>windows</id>
+      <activation>
+        <os><family>windows</family></os>
+      </activation>
+      <properties>
+        <cmd>.cmd</cmd>
+      </properties>
+    </profile>
+  </profiles>
+</project>
--- a/maven-wrapper-plugin/src/it/projects/mavenversion_range/test.properties
+++ b/maven-wrapper-plugin/src/it/projects/mavenversion_range/test.properties
@@ -1,0 +1,18 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+# 
+#   http://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing,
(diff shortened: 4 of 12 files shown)


maven-wrapper-plugin/pom.xml
```diff
@@ -38,7 +38,7 @@ under the License.
   </prerequisites>
 
   <properties>
-    <version.mockito>4.11.0</version.mockito>
+    <version.mockito>5.23.0</version.mockito>
```
giulong: Bumped to use `mockStatic` and `mockConstruction`

maven-wrapper-plugin/pom.xml
```diff
@@ -75,6 +75,12 @@ under the License.
       <artifactId>plexus-io</artifactId>
       <version>3.6.0</version>
     </dependency>
+    <dependency>
+      <groupId>org.codehaus.plexus</groupId>
+      <artifactId>plexus-utils</artifactId>
```
giulong: Added since [MetadataXpp3Reader.read()](https://maven.apache.org/ref/3.3.9/apidocs/org/apache/maven/artifact/repository/metadata/io/xpp3/MetadataXpp3Reader.html#read(java.io.InputStream)) throws `XmlPullParserException` from this package.

cstamas:
-1

Why not use RepositorySystem, do not replicate the logic.

maven-wrapper-plugin/pom.xml
```diff
@@ -113,6 +119,11 @@ under the License.
       <artifactId>junit-jupiter-api</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-params</artifactId>
```
giulong: Added to use `@ParameterizedTest`

cstamas:
You most probably want this instead https://github.com/apache/maven-wrapper/pull/437

cstamas:
Feel free to incorporate that instead.

giulong:
> You most probably want this instead https://github.com/apache/maven-wrapper/pull/437

I was thinking about something like this (currently a draft):

```java
String resolveVersionRange(String version) {
    try {
        Artifact artifact = new DefaultArtifact("org.apache.maven:apache-maven:" + version);
        VersionRangeRequest versionRangeRequest = new VersionRangeRequest();
        versionRangeRequest.setArtifact(artifact);
        versionRangeRequest.setRepositories(session.getCurrentProject().getRemotePluginRepositories());

        return repositorySystem
                .resolveVersionRange(repositorySystemSession, versionRangeRequest)
                .getHighestVersion()
                .toString();
    } catch (VersionRangeResolutionException e) {
        return version;
    }
}
```

Maybe I'm missing something, I'll check your suggestion in the next days

cstamas:
Yes, that looks good as well, but several remarks:
* getHighestVersion may return SNAPSHOT, hence the loop in my code
* version may be just version, so all this is not needed (the logic in my code), but you are right, this mojo is invoked by user, no need to complicate or to be "fast". Simple is better.

Otherwise, yes, this is the right way, no need to fiddle with XML parsing etc.

giulong:
> getHighestVersion may return SNAPSHOT, hence the loop in my code

`VersionRangeRequest` has a [constructor](https://maven.apache.org/resolver/apidocs/org/eclipse/aether/resolution/VersionRangeRequest.html#%3Cinit%3E(org.eclipse.aether.artifact.Artifact,java.util.List,org.eclipse.aether.metadata.Metadata.Nature,java.lang.String)) accepting the nature (snapshot/release/both). Leveraging that one, the method would be:

```java
String resolveMavenVersion(String version) {
    try {
        Artifact artifact = new DefaultArtifact("org.apache.maven:apache-maven:" + version);
        VersionRangeRequest request = new VersionRangeRequest(
                artifact, session.getCurrentProject().getRemotePluginRepositories(), RELEASE, "wrapper"); // <-- here
        
        return repositorySystem
                .resolveVersionRange(repositorySystemSession, request)
                .getHighestVersion()
                .toString();
    } catch (VersionRangeResolutionException e) {
        return version;
    }
}
```

What do you think? Is there anything else to address?

Edit:
that constructor, along with the setter to specify the nature, come with version `2.0.11`, while `maven-core` is bringing version `1.9.27` as transitive dependencies:

```text
[INFO] --- dependency:3.11.0:tree (default-cli) @ maven-wrapper-plugin ---
[INFO] org.apache.maven.plugins:maven-wrapper-plugin:maven-plugin:3.3.5-SNAPSHOT
[...]
[INFO] |  +- org.apache.maven.resolver:maven-resolver-impl:jar:1.9.27:provided
[INFO] |  |  \- org.apache.maven.resolver:maven-resolver-named-locks:jar:1.9.27:provided
[INFO] |  +- org.apache.maven.res`olver:maven-resolver-spi:jar:1.9.27:provided
[INFO] |  +- org.apache.maven.resolver:maven-resolver-util:jar:1.9.27:provided
```

Edit 2:

Fallbacking to your approach for filtering snapshots since we can't use the nature when building the `VersionRangeRequest`:
```java
String resolveMavenVersion(String version) {
    try {
        Artifact artifact = new DefaultArtifact("org.apache.maven:apache-maven:" + version);
        VersionRangeRequest request = new VersionRangeRequest(
                artifact, session.getCurrentProject().getRemotePluginRepositories(), "wrapper");

        List<Version> versions = repositorySystem
                .resolveVersionRange(repositorySystemSession, request)
                .getVersions();
        versions.sort(reverseOrder());

        return versions.stream()
                .map(Version::toString)
                .filter(v -> !artifact.setVersion(v).isSnapshot())
                .findFirst()
                .orElse(version);
    } catch (VersionRangeResolutionException e) {
        return version;
    }
}
```

giulong:
Saw the verify action failing for two kinds of issues:
1. The mockito version that I bumped (from 4 to 5) is compiled with Java 11, leading to [failures on jdk 8](https://github.com/apache/maven-wrapper/actions/runs/30573171750/job/91954868884?pr=436). I reverted it and rewritten the unit tests accordingly.
2. [The new IT is failing on Windows and Maven 4.0.0-rc-6](https://github.com/apache/maven-wrapper/actions/runs/30573171750/job/91954868712?pr=436). This is not yet fixed. My guess is that the version range `maven=[3.0,3.10-rc\)` in [test.properties](https://github.com/apache/maven-wrapper/pull/436/changes#diff-6fe80630453eb5b91748d4e3bb5097c7aa762e106b76b196f8b9937fdcd0c458R18) is not resolved correctly on that Maven-OS combination. Still need to address this one.

cstamas:
While here, does meta versions like `RELEASE` work?

giulong:
As of now, no. We'd need to leverage `VersionRequest` along with `VersionRangeRequest`. Should I also add this one?

giulong:
I confirm the issue on Windows was due to the round brackets escape in the `test.properties`: `maven=[3.0,3.10-rc\)`. Solved by just removing it: `maven=[3.0,3.10-rc)`

That escaping is only needed in bash, where parentheses are interpreted as special chars for subshells.

The weird thing is that the issue didn't arise in Windows with Maven 3, only with Maven 4. The invoker plugin forwards the args in the same way, removing the escape already:

* Maven 3:
  ```sh
  [DEBUG] Executing: cmd.exe /X /C "D:\Giulio\apache-maven-3.9.2\bin\mvn.cmd -B -D maven.repo.local=D:\Giulio\workspace\maven-wrapper\maven-wrapper-plugin\target\local-repo -s D:\Giulio\workspace\maven-wrapper\maven-wrapper-plugin\target\it\interpolated-settings.xml -D maven=[3.0,3.10-rc) org.apache.maven.plugins:maven-wrapper-plugin:3.3.5-SNAPSHOT:wrapper exec:exec"
  ```

* Maven 4:
  ```sh
  [DEBUG] Executing: cmd.exe /X /C "D:\Giulio\apache-maven-4.0.0-rc-6\bin\mvn.cmd -B -D maven.repo.local=D:\Giulio\workspace\maven-wrapper\maven-wrapper-plugin\target\local-repo -s D:\Giulio\workspace\maven-wrapper\maven-wrapper-plugin\target\it\interpolated-settings.xml -D maven=[3.0,3.10-rc) org.apache.maven.plugins:maven-wrapper-plugin:3.3.5-SNAPSHOT:wrapper exec:exec"
  ```

So I don't really get where the issue was. I suspect it's somewhere in the `mvn.cmd`, which is different in the two Maven versions.

Nevertheless, now it should all be fixed. Can you run the verify action again?

giulong:
The issue on Windows and Maven 4 is related to the closing round bracket, which needs to be escaped.
Quoting the value in `maven="[3.0,3.10-rc)"` solves on Windows, but isn't a valid syntax for Linux.

The solution to cover all combinations, as you can see in the [verify action in my fork](https://github.com/giulong/maven-wrapper/actions/runs/32895824380), was to double the IT and run the same assertions:
* [mavenversion_range](https://github.com/apache/maven-wrapper/pull/436/changes#diff-19f58e4079766f1fb04419bd84323c2c439a37befa2e04c4a1a8d5b6029a8482R20) runs with `maven=[3.0,3.10-rc)` on `invoker.os.family = !windows`
* [mavenversion_range_windows](https://github.com/apache/maven-wrapper/pull/436/changes#diff-ed5e3d1ea675d92fd04d80ac5141e45c73f5c1fd5dcedfeac2009b4b30f0bc03R20) runs with `maven="[3.0,3.10-rc)"` on `invoker.os.family = windows`

Can you run the verify action again?

Edit: links

giulong:
Can you guys please run the `verify` action?

PR #443, opened 2026-08-15
Fail when only-script mvnw has no download tool #441

### LINKED ISSUE #441: mvnw silently continues when no download mechanism is available (missing final else)
## Description

In the `mvnw` script generated by maven-wrapper 3.3.4 (`distributionType=only-script`), the download dispatch is a three-branch chain with no final `else`:

```sh
if [ -z "${MVNW_USERNAME-}" ] && command -v wget >/dev/null; then
  wget ... || die "wget: Failed to fetch $distributionUrl"
elif [ -z "${MVNW_USERNAME-}" ] && command -v curl >/dev/null; then
  curl ... || die "curl: Failed to fetch $distributionUrl"
elif set_java_home; then
  # compile and run Downloader.java
fi
```

If all three conditions are false — no `wget`, no `curl`, and `set_java_home` fails (no `JAVA_HOME`, no `java`/`javac` on `PATH`) — the block is a no-op. Execution continues, the SHA-256 validation block is skipped when `distributionSha256Sum` is unset, and the script then attempts to unzip/untar a file that was never downloaded.

The user sees an `unzip`/`tar` failure about a missing or corrupt archive rather than the actual cause, which is that no download mechanism was available.

## Steps to reproduce

Run `./mvnw -v` on a machine where `wget` and `curl` are absent from `PATH`, `JAVA_HOME` is unset, and neither `java` nor `javac` is on `PATH`, with no distribution yet cached under `${MAVEN_USER_HOME}/wrapper/dists`.

## Suggested fix

Add a final `else` that fails with a clear message, and/or assert the archive exists before extraction:

```sh
else
  die "No download mechanism available: install wget or curl, or make a JDK available on PATH / via JAVA_HOME"
fi

[ -f "$TMP_DOWNLOAD_DIR/$distributionUrlName" ] || die "Distribution was not downloaded: $distributionUrl"
```

## Related question

The `wget` and `curl` branches are both gated on `[ -z "${MVNW_USERNAME-}" ]`, so when `MVNW_USERNAME` is set they are skipped even if available, and the script always falls through to the Java downloader. Earlier jar-based wrapper scripts passed credentials directly (`wget --http-user=… --http-password=…`, `curl --user …`). If that is deliberate — because `Downloader.java` centralises auth via `java.net.Authenticator` — it may be worth a comment, since it means a machine with `curl` but no JDK cannot perform an authenticated download.

## Environment

maven-wrapper 3.3.4, `distributionType=only-script`, Maven 3.9.16.

AzazelSensei:

- [ ] Run `mvn verify` to make sure basic checks pass.
- [ ] You have run the integration tests successfully (`mvn -Prun-its verify`).

---

`only-mvnw` tries wget, then curl, then a JDK downloader. There was no
`else`. On a box with none of those, `set_java_home` just returns 1 and
the script keeps going. You get `failed to untar` on a file that was
never fetched.

I added the missing `else` and a check that the archive is actually
there before extract.

Reproduced with a PATH that has `tar` but not wget/curl/java:

- before: `The java/javac command does not exist...` then `failed to untar`
- after: same java note, then `No download mechanism available: ...` and exit 1

The curl branch is unchanged. I didn't run `mvn verify` here; no JDK on
this machine.

Fixes #441


AzazelSensei:
Closing this after 15 days without a comment.

